Issue #2, opened 2018-08-16
Change Gadfly.jl to Plots.jl in plotting functions


xukai92:

trappmartin:
All plotting functions (except for contour plots) are ported and work as illustrated in the demo.

PR #3, opened 2018-08-18
Addressing #2: Change Gadfly.jl to Plots.jl in plotting functions

xukai92:
Addressing #2


test/plot_test.jl
```diff
@@ -0,0 +1,19 @@
+# run this file in interactive mode to pop-up plots (`julia -i plot_test.jl`)
+# NOTE: I don't understand how Plots evals plotting call - I
```
xukai92: @trappmartin See this comment.

trappmartin:
It seems to me that it is suggested to use Recipes for user-defined plotting schemes. See http://docs.juliaplots.org/latest/recipes/. I'll see if such an approach works.

yebai:
@trappmartin Sounds good. Shall we merge this PR first?

PR #5, opened 2018-08-20
Add Project.toml.

hessammehr:
This commit allows the [github.com/TuringLang/Turing.jl/issues/462](https://github.com/TuringLang/Turing.jl/tree/github.com/TuringLang/Turing.jl/issues/462) branch of Turing to compile in Julia 1.0.


trappmartin:
As far as I understand from the discussions on discourse, the official suggestion right now is to still use REQUIRE instead of a Project.toml. I guess this will change but for now I rather stick to REQUIRE instead.

PR #13, opened 2018-09-03
Avoid loading the `Plots` package.

hessammehr:
This does away with the precompilation warnings. Plotting functions
should continue to work as before.

### DIFF
--- a/src/plot.jl
+++ b/src/plot.jl
@@ -1,11 +1,11 @@
 using RecipesBase
 using StatPlots
-import Plots.plot
 #################### Posterior Plot Recipies ####################
-
-@recipe function plot(c::MCMCChain.AbstractChains, ptypes::Vector{Symbol}; 
-                      maxlag = round(Int, 10 * log10(length(c.range))),
-  barbounds = (0, Inf))
+@recipe function f(c::MCMCChain.AbstractChains,
+                   ptypes::Union{Symbol, AbstractArray{Symbol}};
+                   maxlag = round(Int, 10 * log10(length(c.range))),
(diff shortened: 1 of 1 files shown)


hessammehr:
With the following code

```julia
@model toy_model(x) = begin
    p ~ Dirichlet(3, 0.2)
    _,i = findmax(p)
    y ~ Normal(0.0, [1.0, 2.0, 3.0][i])
    x ~ Normal(y, 1.0)
end
I = sample(toy_model(1.0), HMC(10000, 0.4, 10))
x = Chains(I.value, I.range, I.names, I.chains)
meanplot(x)
```

I get

example

trappmartin:
OK, I agree that looks strange. 

I currently do not adjust the plot size in the recipe. I guess this is causing the funny visualisation. Does it look alright if you manually set an appropriate size?

Thanks!

hessammehr:
Much better if you do: 

```julia
size := (800, 1200)
```

example

That said, I was thinking of an [alternative](https://gist.github.com/hessammehr/b5029ebf6e945a3063d1b42e1a00ad9a) using user recipes.

trappmartin:
I tried this in the first place but couldn't figure out how to properly stack them. Your approach looks really neat, could you please open a new PR for it. I'm happy to help, if possible.

That said, the size issue still remains. But we could try to circumvent this by automatically adjusting the plot size or even construct multiple plots.

hessammehr:
Wonderful, @trappmartin! I will open a PR in a second. Would appreciate your help in getting it to work. Agree about pre-calculating plot sizes as a workaround to the size issue.

PR #17, opened 2018-09-19
Refactoring of the plotting functions of MCMCChain

trappmartin:
Improved plotting capacities for MCMCChain.

TODO: 
- add documentation on `colordim`
- check if creating `plot` alias functions is problematic

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,8 @@ script:
   - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
   - julia -e "
         using Pkg;
+        Pkg.add(\"Plots\");
+        Pkg.add(\"StatPlots\");
         Pkg.clone(pwd(), \"MCMCChain\");
         Pkg.build(\"MCMCChain\");
         Pkg.test(\"MCMCChain\"; coverage=true)
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,26 +1,30 @@
 language: julia
+
 os:
   - linux
-  - osx
+#  - osx
+
 julia:
   - 1.0
   - nightly
+
 matrix:
   allow_failures:
     - julia: nightly
   fast_finish: true
+
 notifications:
   email: false
-script:
-  - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
-  - julia -e "
-        using Pkg;
-        Pkg.add(\"Plots\");
-        Pkg.add(\"StatPlots\");
-        Pkg.clone(pwd(), \"MCMCChain\");
-        Pkg.build(\"MCMCChain\");
-        Pkg.test(\"MCMCChain\"; coverage=true)
-      "
+
+#script:
+#  - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
+#  - julia -e "
+#        using Pkg;
+#        Pkg.clone(pwd(), \"MCMCChain\");
+#        Pkg.build(\"MCMCChain\");
+#        Pkg.test(\"MCMCChain\"; coverage=true)
+#      "
+
 after_success:
   - julia -e 'using Pkg; cd(Pkg.dir("MCMCChain")); Pkg.add("Coverage"); using Coverage; Coveralls.submit(process_folder())'
 
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ chn = Chains(val);
 
 # visualize the MCMC simulation results 
 p1 = plot(chn)
-p1 = plot(chn, colordim = :parameter)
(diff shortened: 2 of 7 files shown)


.travis.yml
```diff
@@ -1,24 +1,30 @@
...
-script:
-  - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
-  - julia -e "
-        using Pkg;
-        Pkg.clone(pwd(), \"MCMCChain\");
-        Pkg.build(\"MCMCChain\");
-        Pkg.test(\"MCMCChain\"; coverage=true)
-      "
+
+#script:
```
yebai: any reason for removing Travis tests?

README.md
```diff
@@ -23,15 +23,15 @@ val = hcat(val, rand(1:2, n_iter, 1, n_chain))
 # construct a Chains object
 chn = Chains(val)
 
-# visualize a density plot / histogram plot, autocorrelation plot and a running average plot
-plot(chn, [:mixeddensity, :autocor, :mean])
+# visualize the MCMC simulation results 
+plot(chn)
```
yebai: This new default plot looks nice. Do we still have support for the old syntax `plot(chn, [:mixeddensity, :autocor, :mean])`?

README.md
```diff
@@ -85,34 +85,31 @@ rafterydiag(c::AbstractChains; q=0.025, r=0.005, s=0.95, eps=0.001)
 ### Plotting
 ```julia
 # construct a plot
-plot(c::AbstractChains, type::Symbol)
+plot(c::AbstractChains; ptypes = [TracePlot, MixedDensityPlot])
```
yebai: To help downstream packages, it might be helpful to add an alias function to keep supporting the old syntax: `plot(c::AbstractChains, :density)`?

src/plot.jl
```diff
@@ -1,79 +1,83 @@
...
+@recipe function f(p::MixedDensityPlot, barbounds = (0, Inf))
+  c, i = p.args
+  discrete = MCMCChain.indiscretesupport(c, barbounds)
+  seriestype := discrete[i] ? :histogram : :density
+  xaxis := "Sample value"
+  yaxis := discrete[i] ? "Frequency" : "Density"
+  c.value[:, i, :]
+end
+
+@recipe function f(c::MCMCChain.AbstractChains;
```
yebai: Perhaps tweak the style a bit (e.g. reduce title font size, axis font size)? The [one](https://github.com/TuringLang/MCMCChain.jl/issues/15#issuecomment-420339484) from PyMC3 looks quite nice actually.

yebai:
Thanks, @trappmartin. It looks great! I left some minor comments above.

.travis.yml
trappmartin: I uncommented these lines as it is sufficient and probably better to run the default Travis script for Julia packages. We do not need any specialized code here and using the default removes the burden of maintaining the Travis file.

README.md
trappmartin: Unfortunately not. We would need to specify a plot(...) function which would create a naming clash with Plots and to resolve this we would need to import and depend on Plots. This solution has a more clumsy syntax but frees us from Plots and StatPlots.

src/plot.jl
trappmartin: Unfortunately, we cannot set font types and sizes without depending on Plots. However, there is a nice package called PlotThemes which automatically sets several parameters according to predefined themes. I'll refer to this package if the user wants to improve the look of the plots.

README.md
trappmartin: I was wrong. In fact, `Plots.jl` depends on `RecipesBase.jl` which exports `plot` and `plot!`. To support the old syntax we have to provide alias functions and export them. I'll push the code asap.

README.md
trappmartin: Same here, I'll provide alias functions that support the old syntax. The functions will raise a warning indicating that one should use the new syntax.

src/plot.jl
yebai: Perhaps this can be solved using `Requires`, see 

https://github.com/MikeInnes/Requires.jl

trappmartin:
Here are some example plots of the current version with `theme(:ggplot2)` as selected theme.

Running `plot(chn)` constructs the following default plot.
p1

Running `plot(chn, colordim = :parameter)` constructs the following default plot.
p2

yebai:
> Here are some example plots of the current version with theme(:ggplot2) as selected theme.

Looks nice!

trappmartin:
@yebai this PR is finished from my side and can be squashed and merged into the master.

Issue #16, opened 2018-09-17
Allow missing values


trappmartin:
The current implementation does not allow missing values. In favour of more flexible models, we should allow the Chains type to work with missing values.

PR #18, opened 2018-09-19
Refactoring of the Chain type to allow for missing values

trappmartin:
This PR refactors the Chain type in MCMCChain to allow for missing values and additionally introduces unique symbols which can be used to uniquely index variable.

Related issues #16, TuringLang/Turing.jl#474, TuringLang/Turing.jl#484, TuringLang/Turing.jl#370


yebai:
Thanks, @trappmartin.

PR #19, opened 2018-09-23
RFC: Reimplement plots

mkborregaard:
As discussed on Slack. Here's the rationale (copied from the discussion). 

Right now the code is elegant and makes full use of the recipes API, but it’s just a little at odds with normal Plots logic. In two respects mainly: 

1. You have `densityplot` and `histogramplot` names, but in fact they’re just `density` and `histogram` that accepts a `Chain` (and adds some labels). The Plot idea is to use the same logic throughout, so you won’t have to remember so many arbitrary things, so ideally you should just write `histogram(myChain)`. But that’s a little tricky as density and histogram are `series recipes` and you’ve implemented everything as `user recipes`.

2. You have the concept of `ptype`, which is exactly what we in Plots terminology call `seriestype`. So I’d suggest reimplementing using the `seriestype` interface. It’s tricky in practice because you want each chain to have a different color, and seriestypes don’t normally do that. 

3. One way of doing this is slightly unelegant by simply querying the seriestype argument and then using `if...else` loops. I've used object dispatch and recipes instead where possible, for code cleanliness, but it turned out to not be completely possible for AutocorPlot. I'm not really happy about that, but there you are. It could all be reimplemented as the if-else block instead.

3. I've also cleaned duplications from the code.

cc @trappmartin

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,13 +3,13 @@
 [![Build status](https://ci.appveyor.com/api/projects/status/1av8osv0099nqw8m/branch/master?svg=true)](https://ci.appveyor.com/project/trappmartin/mcmcchain-jl/branch/master)
 [![Coverage Status](https://coveralls.io/repos/github/TuringLang/MCMCChain.jl/badge.svg?branch=master)](https://coveralls.io/github/TuringLang/MCMCChain.jl?branch=master)
 
-Implementation of Julia types for summarizing MCMC simulations and utility functions for diagnostics and visualizations. 
+Implementation of Julia types for summarizing MCMC simulations and utility functions for diagnostics and visualizations.
 
 ## Example
 The following simple example illustrates how to use Chain to visually summarize a MCMC simulation:
 ```julia
 using MCMCChain
-using Plots, StatPlots
+using StatPlots
 
 theme(:ggplot2);
 
@@ -25,7 +25,7 @@ val = hcat(val, rand(1:2, n_iter, 1, n_chain));
 # construct a Chains object
 chn = Chains(val);
 
-# visualize the MCMC simulation results 
+# visualize the MCMC simulation results
 p1 = plot(chn)
 p2 = plot(chn, colordim = :parameter)
 
@@ -41,21 +41,21 @@ Summarize parameters | Summarize chains
 `plot(chn; colordim = :chain)` | `plot(chn; colordim = :parameter)`
 ![p1](https://user-images.githubusercontent.com/7974003/45822242-f0009180-bce2-11e8-8fa0-a97c8732400f.png)  |  ![p2](https://user-images.githubusercontent.com/7974003/45822249-f131be80-bce2-11e8-8dd3-42db7d58abd9.png)
 
- 
+
 
 ## Manual
 ### Chains type
 ```julia
 # construction of a Chains object
 Chains(iterations::Int, params::Int;
-    start = 1, thin = 1, chains = 1, 
+    start = 1, thin = 1, chains = 1,
     names = String[])
 
-# construction of a Chains object using an 
+# construction of a Chains object using an
 # iteration * params * chains
 # array (values).
-Chains(values::Array{T, 3}; 
-    start = 1, thin = 1, chains = 1, 
+Chains(values::Array{T, 3};
+    start = 1, thin = 1, chains = 1,
     names = String[])
 
 # Indexing a Chains object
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -3,13 +3,13 @@
...
 [![Coverage Status](https://coveralls.io/repos/github/TuringLang/MCMCChain.jl/badge.svg?branch=master)](https://coveralls.io/github/TuringLang/MCMCChain.jl?branch=master)
 
-Implementation of Julia types for summarizing MCMC simulations and utility functions for diagnostics and visualizations. 
+Implementation of Julia types for summarizing MCMC simulations and utility functions for diagnostics and visualizations.
 
 ## Example
 The following simple example illustrates how to use Chain to visually summarize a MCMC simulation:
 ```julia
 using MCMCChain
-using Plots, StatPlots
```
mkborregaard: At the moment you don't need to load Plots separately when loading StatPlots. That may change in the future.

README.md
```diff
@@ -41,21 +41,21 @@ Summarize parameters | Summarize chains
...
 
- 
+
 
 ## Manual
 ### Chains type
 ```julia
 # construction of a Chains object
 Chains(iterations::Int, params::Int;
-    start = 1, thin = 1, chains = 1,
```
mkborregaard: sorry, the cut whitespace is done by my text editor

README.md
```diff
@@ -95,31 +95,44 @@ rafterydiag(c::AbstractChains; q=0.025, r=0.005, s=0.95, eps=0.001)
 ### Plotting
 ```julia
 # construct a plot
-plot(c::AbstractChains; ptypes = [TracePlot, MixedDensityPlot])
+plot(c::AbstractChains, seriestype = (:traceplot, :mixeddensity))
```
mkborregaard: Updated the syntax here. Tuples work better than a Vector for seriestype

src/plot.jl
```diff
@@ -1,164 +1,110 @@
...
-
-@recipe function f(p::MeanPlot; colordim = :chain)
-    c, i = p.args
-    seriestype := :line
-    xaxis := "Iteration"
-    yaxis := "Mean"
-    
-    @assert colordim ∈ colorparamters
+
+@shorthands meanplot
```
mkborregaard: `shorthands` creates new series recipes, whereas `userplot` creates new user recipes.

src/plot.jl
```diff
@@ -1,164 +1,110 @@
...
+@shorthands autocorplot
+@shorthands mixeddensity
+@shorthands traceplot
+
+struct _TracePlot; c; val; end
```
mkborregaard: For dispatch later, I've added these manually.

src/plot.jl
```diff
@@ -1,164 +1,110 @@
...
+struct _MeanPlot; c; val;  end
+struct _DensityPlot; c; val;  end
+struct _HistogramPlot; c; val;  end
+struct _AutocorPlot; lags; val;  end
+
+# define alias functions for old syntax
+const translationdict = Dict(
```
mkborregaard: moved up (and made const)

src/plot.jl
```diff
@@ -1,164 +1,110 @@
...
+        st = (discrete[i] && colordim == :chain) ? :histogram : :density
+        seriestype := st
+    end
 
-    @assert colordim ∈ colorparamters
-    if colordim == :parameter
-        title := "Chain $(c.chains[i])"
-        labels := c.names
-        lags, ac.value[:, :, i]'
+    if st == :autocorplot
```
mkborregaard: unfortunately special casing is needed for autocorplot, it's not so elegant but hey

src/plot.jl
```diff
@@ -1,164 +1,110 @@
...
+        ac = MCMCChain.autocor(c, lags=collect(lags))
+        val = colordim == :parameter ? ac.value[:, :, i]' : ac.value[i, :, :]
+        _AutocorPlot(lags, val)
+    elseif st ∈ supportedplots
+        translationdict[st](c, val)
     else
-        title := c.names[i]
-        labels := map(k -> "Chain $(c.chains[k])", 1:size(c)[3])
-        lags, ac.value[i, :, :]
+        c.range, val
```
mkborregaard: allows other inbuilt seriestypes to be used as well. E.g. `:scatter`

src/plot.jl
```diff
@@ -168,69 +114,21 @@ end
...
                 colordim := colordim
-                ptype([c, i])
+                seriestype := ptype
+                c, i
             end
         end
     end
 end
 
-"""
```
mkborregaard: This whole function has been replaced by line 100

src/plot.jl
```diff
@@ -168,69 +114,21 @@ end
...
-        @series begin
-            subplot := indices[i, 1]
-            colordim := colordim
-            ptype([c, i])
-        end
-    end
-end
-
-# define alias functions for old syntax
-translationdict = Dict(
```
mkborregaard: moved up

mkborregaard:
There's a local test failure on my machine that appears unrelated.

mkborregaard:
There's a GR failure on Travis, though. Maybe set the backend to `plotly` for testing?

trappmartin:
Thanks a lot for the efforts! The PR looks really good. I'll look into changing the backend for the tests in the next days and merge your PR.

trappmartin:
Btw: I think the breaking test should be fixed by the last commit of @yebai.

yebai:
> Btw: I think the breaking test should be fixed by the last commit of @yebai.

That's right. Sorry for breaking tests - I removed the return value from the `describe` function, which some tests obviously depend on. These tests are fixed now.

PR #27, opened 2019-01-17
Corner plot - close #26

### LINKED ISSUE #26: Add cornerplot function
It would be nice to have a quick function that accepts a chain and a variable set and returns a corner plot. We currently do something like this in the [logistic regression tutorial](http://turing.ml/tutorials/2-logisticregression/) for Turing, but it was outside the library. 

In that case, we did something like this:

```julia
# The labels to use.
l = [:student, :balance, :income]

# Extract the parameters we want to plot.
w1 = chain[:student]
w2 = chain[:balance]
w3 = chain[:income]

# Show the corner plot.
cornerplot(hcat(w1, w2, w3), compact=true, labels = l)
```

It would be nice to be able to call `cornerplot(chain, [:var1, :var2, :var3])` and get a plot back. @mkborregaard would you be willing to help on this one?

mkborregaard:
This works now in the sense that
```julia
using Turing, StatPlots, MCMCChain
@model f(x) = begin
          A ~ Normal(0,1)
          B ~ Normal(0,1)
          C ~ Normal(0,1)
          x ~ Normal(0,1)
       end
chain = sample(f(1), HMC(500,0.01,1))
chaincorner(chain, [:A, :B, :C])
```
gives
skaermbillede 2019-01-17 kl 21 29 15

But, I really think we should align the syntax with the rest of the plotting syntax you have here. I can't 100% see the design of that so would be good with input. @trappmartin @cpfiffer 

Also, this requires a change in StatPlots so will require StatPlots 0.9, and will only work on julia-1.0+ 
That is a common issue with advanced recipes.

Finally, shouldn't the default size of those points in cornerplot not be smaller? :-O

### DIFF
--- a/src/plot.jl
+++ b/src/plot.jl
@@ -22,6 +22,8 @@ const translationdict = Dict(
 
 const supportedplots = push!(collect(keys(translationdict)), :mixeddensity)
 
+@recipe f(c::AbstractChains, s::Symbol) = c, indexin([s], keys(c))
+
 @recipe function f(c::AbstractChains, i::Int; colordim = :chain, barbounds = (0, Inf), maxlag = nothing)
     st = get(plotattributes, :seriestype, :traceplot)
 
@@ -95,6 +97,7 @@ end
 end
 
 @recipe function f(c::MCMCChain.AbstractChains;
+                   parameters = Symbol[],
                    width = 500,
                    height = 250,
                    colordim = :chain
@@ -127,7 +130,8 @@ end
             end
         end
     else
-        Corner(c)
+        params = isempty(parameters) ? Symbol.(keys(c)) : Symbol.(parameters)
+        Corner(c, params)
     end
 end
 
@@ -143,12 +147,14 @@ function plot(c::AbstractChains, psym::Symbol; args...)
     return plot(c; seriestype = psym, args...)
 end
 
-struct Corner; c; end
+struct Corner
+    c
+    parameters
+end
 
 @recipe function f(corner::Corner)
-    syms = keys(corner.c)
-    label --> permutedims(syms)
+    label --> permutedims(corner.parameters)
     compact --> true
-    size --> (800, 800)
-    RecipesBase.recipetype(:cornerplot, reduce(hcat, corner.c[s] for s in Symbol.(syms)))
+    size --> (600, 600)
+    RecipesBase.recipetype(:cornerplot, reduce(hcat, corner.c[s] for s in corner.parameters))
 end
--- a/src/plot.jl
+++ b/src/plot.jl
@@ -125,7 +125,6 @@ end
         for (j, ptype) in enumerate(ptypes)
             for (i, par) in enumerate(parameters)
                 @series begin
-                    @show parameters
                     subplot := indices[i, j]
                     colordim := colordim
                     seriestype := ptype
--- a/src/plot.jl
+++ b/src/plot.jl
@@ -137,3 +137,12 @@ function plot(c::AbstractChains, psym::Symbol; args...)
     @warn "This syntax is deprecated, please use plot(c, seriestype = $(translationdict[psym])) instead"
     return plot(c; seriestype = psym, args...)
 end
+
+@userplot ChainCorner
+@recipe function f(cs::ChainCorner)
+    chain, syms = cs.args
+
+    label --> permutedims(syms)
+    compact --> true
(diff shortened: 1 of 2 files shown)


mkborregaard:
Changed the syntax to simply
```
plot(mychain, seriestype = :corner)
```
will currently not work so well because `keys` gives too many names, but that should be fixed in `keys` (I think).
Requires the most recent release of StatPlots (0.9)

cpfiffer:
Does the way these recipes are specified allow for us to add a keyword argument to `plot`? Can we support something like `plot(chain, seriestype = :corner, params = [:A, :B, :C])`?

Also, thank you for your help, @mkborregaard! It would take me what I suspect is a very long time to assemble this.

mkborregaard:
Yes it does allow that - it's only a question of choosing the best coherent syntax. `plot(chain, [:A, :B, :C], seriestype = :corner)` would be another possibility. In both cases it would be relevant to think about whether these are useful for all the other seriestypes as well - e.g. if `colordim = :chain` would it make sense to also select parameters from the other seriestypes?
btw `parameter` is a much better keyword than `param` imho - the short lingo-forms remind me of R 🙀 :-)

cpfiffer:
Yes, `plot(chain, [:A, :B, :C], seriestype = :corner)` should work in all cases. If nothing is provided for the symbol vector, then it should default to using `keys(chain)` for the values to plot. That'd also make all this more future-proof. So I suppose we don't actually need a keyword argument for symbols.

mkborregaard:
I've updated the syntax to
```
plot(chain, parameters = [:A, :B, :C], seriestype = :corner)
# or simply
corner(chain, parameters = [:A, :B, :C])
```
It works now so can be merged if that's the final syntax. 
I think this could be improved to also allow the other plot types to accept these keyword args, to set colordim automatically to :chain when `parameters` is given, etc.

I've for starters now also added the syntax
```
plot(chain, :A, seriestype = A)
```
for `A in (traceplot, meanplot, density, histogram, autocorplot)`

Also added a code snippet to the readme, and removed the earlier verbose deprecations.

```

**Comment by mkborregaard** [comment]
Haha sorry just saw that you'd commented you did NOT want a keyword argument for parameters. Sorry! Fixing.

**Comment by mkborregaard** [comment]
OK, updated.
All the seriestypes now take a symbol or vector of symbols (which will be interpreted as parameter names and require colordim to be :chain), an Int (it already did) or Vector{Int}, which will be interpreted as indices into `keys(chain)`. The single-argument form is not touched.

Not sure the code is so clean anymore :-O There's a lot of checking, as corner is not compatible with the other seriestypes etc. Not really sure how to clean that up nicely.

Also note that on my computer the `corner` code **segfaults** if some keys are selected that are not the parameters. Really - not sure what's going on there.

**Comment by yebai** [comment]
Excellent work. Ready to merge whenever you guys are happy!

**Comment by mkborregaard** [comment]
Great. IMHO it's good to merge :shipit:
But you may want to play around with this for real models from the wild, I've only played with the toy model above.

File: src/plot.jl [hunk 1]
```diff
@@ -93,47 +96,57 @@ end
...
+
+    if !(:corner ∈ ptypes)
+        layout := (N, ntypes)
+        size --> (ntypes*width, N*height)
+        indices = reshape(1:N*ntypes, ntypes, N)'
+        legend --> false
+        for (j, ptype) in enumerate(ptypes)
+            for (i, par) in enumerate(parameters)
+                @series begin
+                    @show parameters
```
benelsen: I believe this line still needs to go or `plot(chain)` will be a bit too verbose.

benelsen:
I tested it on https://github.com/StanJulia/StatisticalRethinking.jl/blob/master/notebooks/12/m12.6.ipynb and after realising this is in fact the compact version of the StatPlots' `cornerplot`, it looks right to me 
screenshot from 2019-01-22 22-04-19

src/plot.jl
mkborregaard: :joy: ouch - good catch!

mkborregaard: fixed

mkborregaard:
Great - btw I plan to change the default colorscheme in StatPlots. I'm not sure people realize this, but the color indicates the strength of the correlation. By using a different colormap, we now have blue for positive correlations, red for negative, yellow for neutral. Comments? This is unrelated to this PR.

skaermbillede 2019-01-22 kl 22 49 27

cpfiffer:
I did not realize that the colors indicated correlation! I'm fond of the new colors.

PR #31, opened 2019-01-31
Fixed isnothing not defined error

cpfiffer:
I was getting an undefined error for the `isnothing` function in `corner(chain, syms)` on Julia 1.0, so I changed it to an anonymous function. Anyone see any obvious issues here?

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -4,7 +4,7 @@ using Pkg
 # add packages
 to_add = [
     PackageSpec(name="Plots"),
-    PackageSpec(name="StatPlots"),
+    PackageSpec(name="StatsPlots"),
 ]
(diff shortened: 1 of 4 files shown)


test/runtests.jl
```diff
@@ -4,7 +4,7 @@ using Pkg
 # add packages
 to_add = [
     PackageSpec(name="Plots"),
-    PackageSpec(name="StatPlots"),
+    PackageSpec(name="StatsPlots"),
```
mkborregaard: Hilariously you spelled it wrong in the commit message :-)

cpfiffer: Hey, I got it right where it counts. Good eye!

PR #32, opened 2019-02-10
Serialization/Deserialization Support

cpfiffer:
Fixes support for serialization and deserialization of `AbstractChains` objects, and adds a test to make sure that everything's working right.

I've imported `Serialization`, do I need to add it to the REQUIRE file too?


goedman:
No, it is a stdlib. It only was introduced in Julia 1.1.

Issue #37, opened 2019-02-23
MCMCChains.jl exports "plot"


dcarrera:
The MCMCChain.jl package adds a "plot" function to the global namespace based on Plots.jl. This creates a problem for those of us who do not use Plots.jl as our preferred plotting package. I think that an MCMC code should not be polluting the global namespace with plotting functions and causing issues with people who do not share the developers' preference of plotting package. An MCMC code is not a plotting module. It should be possible (and default) to get the MCMC features alone.

Would it be possible for MCMCChain.jl to explicitly only overwrite `Plots.plot` or maybe check whether `Plots` is even being used?

cpfiffer:
I've opened #39 to address this, which does not load any plotting functionality unless a user calls `using StatsPlots` or `using Plots`. 

Please be aware that this package is currently undergoing a name change (from `MCMCChain` to `MCMCChains`) and some fairly significant refactoring on the back end. After #39 is through, the fix will show up in a new package `MCMCChains`, not `MCMCChain`. You can see the progress of the package registration issue [here](https://github.com/JuliaLang/METADATA.jl/pull/21638).

mkborregaard:
I agree that the package should never pollute the global namespace with plotting functions, but this is easy to fix. See my comment on the PR.

cpfiffer:
Closed by #39.

PR #39, opened 2019-02-23
Remove plot loading by default

cpfiffer:
See #37. MCMCChains will now only export (and load) `plot` if `Plots` or `StatsPlots` is loaded first. Anyone see any obvious issues with this implementation?

### DIFF
--- a/src/MCMCChains.jl
+++ b/src/MCMCChains.jl
@@ -8,7 +8,9 @@ import Serialization: serialize, deserialize
 import Base: sort, range, names
 import Statistics: cor
 
-using Requires
+using RecipesBase
+import RecipesBase: plot
+
 using Serialization
 using Distributions
 using SpecialFunctions
@@ -53,9 +55,6 @@ include("heideldiag.jl")
 include("mcse.jl")
 include("rafterydiag.jl")
 include("stats.jl")
-
-function __init__()
-    @require Plots="91a5bcdd-55d7-5caf-9e0b-520d859cae80" @eval include("plot.jl")
-end
+include("plot.jl")
 
 end # module
--- a/src/MCMCChains.jl
+++ b/src/MCMCChains.jl
@@ -8,9 +8,7 @@ import Serialization: serialize, deserialize
 import Base: sort, range, names
 import Statistics: cor
 
-using RecipesBase
-import RecipesBase: plot
-
+using Requires
 using Serialization
 using Distributions
 using SpecialFunctions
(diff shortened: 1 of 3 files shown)


goedman:
Should we in case either of these are loaded ‘import Plots: plots’ and add a plot() method for MCMCChains.Chains?

In Pkg3 land I wonder if testing for Plots or StatsPlots is the intended way to go. Maybe the right solution is a separate MCMCChainsPlots package?

src/MCMCChains.jl
```diff
@@ -8,17 +8,14 @@ import Serialization: serialize, deserialize
...
-
+using Requires
 using Serialization
 using Distributions
 using SpecialFunctions
 using AxisArrays
 const axes = Base.axes
 
 export Chains, getindex, setindex!, chains
-export plot, traceplot, meanplot, densityplot, histogramplot, mixeddensityplot, autcorplot
```
mkborregaard: all you have to do is delete this line - these should not be exported, they will become automagically available when Plots is loaded (you can also see that there is a typo in `autcorplot` which has of course had no effect)

mkborregaard:
No reason for glue packages, Requires or anything else - that is the beauty of recipes. The philosophy is to never cause any problems for people using other plotting packages

mkborregaard:
Fixes #37

goedman:
Michael, care (if possible in not too many words) to describe how PlotRecipes achieves this. I tried to grasp this way back when PlotRecipes were introduced but got lost.

In StanJulia I use overloading of a function (defined in CmdStan) with methods defined in packages such as StanMCMCChain. PlotRecipes must go a step beyond this (no `import Package: function name`).

cpfiffer:
Thank you, @mkborregaard! Saving the day, as always. I'm merging this for the moment.

mkborregaard:
@goedman it works by creating a method for a function called `RecipesBase.apply_recipe` that takes in the recipe syntax and generates a Dict holding all attributes of the plot. The `Plots.plot` function calls this function, dispatching on the plotted object (here `AbstractChain`), when creating the `Plots.Plot` struct (the plot is created when calling `display` on that object, which the REPL does implicitly).

goedman:
Thanks @mkborregaard. that helps to position it. For my simple cases in StanJulia method overloading is fine, but for a _planned_ function like [shade() in StatisticalRethinking](https://statisticalrethinkingjulia.github.io/StanModels.jl/latest/05/m5.1s/) I will attempt to write a plot recipe.

Issue #40, opened 2019-02-23
Serialization tests failing after type redefinition


cpfiffer:
I just changed the type definition of `Chains.value` from 

```julia
value::AxisArray
```

to

```julia
value::AxisArray{Union{Missing,A},3}
```

The stored chain binary at `test/sections_m10.4s.jls` is now no longer of the same size -- @goedman would you mind if I wrote a slightly more robust test using random numbers?

goedman:
That would be great, this was intended as a placeholder until we can create an MCMCChains.Chains object from Turing and pushing the current sections implementation by creating a (useful) third section.

cpfiffer:
Ah, okay. That sounds like a plan. I'll leave this open until the Turing integration goes through, and then I'll revisit the testing suite a little more.

PR #41, opened 2019-02-24
Slightly more replicable serialization tests

cpfiffer:
This PR simplifies the serialization tests a little bit (see #40). @goedman, is there anything here you'd like to see that I have removed?


goedman:
This is fine. The other component (a new section like :pooled) belongs in sections_tests anyway.

PR #42, opened 2019-02-24
Added NamedTuple functionality to Chains objects

cpfiffer:
This PR changes the definition of a `Chains` object to 

```julia
struct Chains{A, T, K<:NamedTuple, L<:NamedTuple} <: AbstractChains
    value::AxisArray{Union{Missing,A},3}
    logevidence::T
    name_map::K
    info::L
end
```

Previously, `name_map` and `info` were of type `Dict`. This makes `Chains` object significantly more type stable. A new external function `setinfo(c::Chains, n::NamedTuple)` was added, which returns a new chain with `n` stored in `c.info`.

Additionally, I added some docstrings and did a little bit of tidying up.

@goedman do you see any conflicts here with anything on your end?

### DIFF
--- a/src/MCMCChains.jl
+++ b/src/MCMCChains.jl
@@ -17,7 +17,7 @@ using SpecialFunctions
 using AxisArrays
 const axes = Base.axes
 
-export Chains, getindex, setindex!, chains
+export Chains, getindex, setindex!, chains, setinfo
 export describe
 
 # export diagnostics functions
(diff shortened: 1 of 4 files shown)


mohamed82008:
I would make the first field's type a type parameter as well, and remove the `<: NamedTuple` to allow for generic structs.

cpfiffer:
Okay, can do. I'm not sure what the benefit is to removing the type constraints there, though. A lot of the code relies on the type of `info` or `name_map`, and I would lean towards enforcing some very specific behavior for those two fields. `name_map` in particular is heavily dependent on being the type that it is. Is there a better way to do that than how I'm doing it now?

mohamed82008:
Hmm, it's fine to keep the `<: NamedTuple` if your code heavily relies on it.

goedman:
I’ll play around with it tonight.

goedman:
I don't see major conflicts.

PR #48, opened 2019-03-06
Updated serialization tests to use Turing

cpfiffer:
Updates the serialization/deserialization tests to use a Turing model to make sure everything's kosher. This also tests to make sure that anything in the `Chains.info` field makes it through by trying to resume sampling from a deserialized chain.


cpfiffer:
Addresses #40, I should mention.

PR #50, opened 2019-03-07
make recipe labels overrideable

mkborregaard:

PR #51, opened 2019-03-07
Label fixes

cpfiffer:
Minor update to make #50 work a little better. I've also changed the default chain name from "Chain1" to just the integer 1. Makes everything look prettier.

### DIFF
--- a/src/chains.jl
+++ b/src/chains.jl
@@ -47,7 +47,7 @@ function Chains(val::AbstractArray{A,3},
     axvals = [
         Base.range(start, step=thin, length=size(val, 1)),
         parameter_names,
-        map(i->i, 1:size(val, 3)),
+        collect(1:size(val, 3)),
     ]
 
     if length(keys(name_map)) == 1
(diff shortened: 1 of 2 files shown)


src/chains.jl
```diff
@@ -47,7 +47,7 @@ function Chains(val::AbstractArray{A,3},
     axvals = [
         Base.range(start, step=thin, length=size(val, 1)),
         parameter_names,
-        map(i->Symbol("Chain$i"), 1:size(val, 3)),
+        map(i->i, 1:size(val, 3)),
```
mkborregaard: Shouldn't that be `Symbol.(1:size(val, 3))` or am I missing something?

cpfiffer: Well, there's no real need to have chain names be symbols. I was lazy and left the `map` call in. I've pushed a modification to make this just `collect(1:size(val, 3))` instead.

mkborregaard: great!

Issue #53, opened 2019-03-08
Unify indexing convention [RFC]


cpfiffer:
Currently, indexing a `Chains` object can return different object types depending on how you index. 

```julia
# Returns a Chains object with P[1] as its only parameter
chn[:, "P[1]", :] 

# Returns an array of the relevant parameter(s)
chn[:P]
```

I'm not sure I like this. In Turing, `chn[:P]` was honestly much better. If I had a vector of parameters, `chn[:P]` would return a vector of vectors `[ P[1], P[2], P[3] ] `.  In MCMCChains v0.3, you get a flat array where it can be difficult to assign columns to each individual `[ P[1], P[2], P[3] ] `. This behavior is kind of weird though and I think it's better served as a separate, explicit function.

I would like to change indexing so that these two indexing patterns are more consistent. The most obvious choice is that `chn[:P]` should do the same thing as `chn[:,["P[1]", "P[2]", "P3]"],:]` which would return a new `Chains` object containing only `"P[1]", "P[2]", "P[3]"`. 

Would people be okay with that? I would also like to provide an additional function like `get_params` or something, where `get_params(chn, :P)` which would return a tuple looking like `(P[1], P[2], P[3])`. This way, the parameter vectors can be packaged in a more accessible way to users who want to work that way (which I do).

goedman:
Definitely a bit more work for CmdStan & DynamicHMC (not sure about Mamba yet), but I've never liked the current solution in CmdStan when effectively an array is returned.

As you probably have noticed, one of the areas I sometimes get lost in MCMCChains is when to create a new Chains object myself or have MCMCChains handle it for me, e.g. ```chn[:P]```.

I think having users just deal with ```chn[:P]``` and ```chn[draws, :, :]``` would have my preference.

cpfiffer:
What do you mean deal with them? We should or should not unify the behavior between `chn[:P]` and `chn[:,"P[1]",:]`?

goedman:
I would vote to unify them.

I meant folks (hopefully everybody!) wanting to use MCMCChains.Chains not generated by Turing initially will have to create such a chain but after that it would be great if they just can select parameters as you suggest, chns[:P] (and not go the route I often chose to create a new Chains object themselves).

By the way, in your example, P[1], P[2] are just column/parameter names? I mean P is not really a matrix, is it?

cpfiffer:
Yeah, they are columns. P wasn't a matrix. 

[...]

PR #57, opened 2019-03-09
[RFC] Index functionality change

cpfiffer:
Addresses #53 by forcing `chn[:D]` and `chn[:, "D", :]` to both return `Chains` objects. This is a request for comments, so please let me know if you don't like this change or want to add/drop things.

Example:

```julia
chn[:, "D", :]
```
```
Object of type Chains, with data of type 500×1×1 Array{Union{Missing, Float64},3}

Log evidence      = 0.0
Iterations        = 1:500
Thinning interval = 1
Chains            = Chain1
Samples per chain = 500
parameters        = D

parameters
   Mean    SD   Naive SE  MCSE     ESS  
D 0.4872 0.2823   0.0126 0.0139 412.6237
```

and 

```julia
chn[:D]
```
```
Object of type Chains, with data of type 500×1×1 Array{Union{Missing, Float64},3}

Log evidence      = 0.0
Iterations        = 1:500
Thinning interval = 1
Chains            = Chain1
Samples per chain = 500
parameters        = D

parameters
   Mean    SD   Naive SE  MCSE     ESS  
D 0.4872 0.2823   0.0126 0.0139 412.6237
```

Users can still extract numerical data if they so wish using `chn[:D].value`. 

I have also added a fairly handy function `get`, which functions very similarly to how `chn[:D]` used to function on the Turing side, but with some extra bells and whistles. I became fed up working with the `AxisArrays` based indexing while trying to update Turing's tutorials, and decided it would be a good idea to provide some support for significantly simpler indexing. `get(chn, :P)` returns a `NamedTuple` which is very easy to work with (in my opinion).

Example:

```julia
val = rand(500, 5, 1)
chn = Chains(val, ["P[1]", "P[2]", "P[3]", "D", "E"]);

x = get(chn, :P)
```

Here's what `x` looks like:

```julia
(P = (Union{Missing, Float64}[0.349592; 0.671365; … ; 0.319421; 0.298899], Union{Missing, Float64}[0.757884; 0.720212; … ; 0.471339; 0.5381], Union{Missing, Float64}[0.240626; 0.987814; … ; 0.980652; 0.149805]),)
```

You can access each of the `P[. . .]` variables by indexing, using `x.P[1]`, `x.P[2]`, or `x.P[3]`. This makes it _very_ easy to work with high dimensional vectors put out by Turing. Hopefully this is also helpful on the StanJulia side.

`get` also accepts vectors of things to get, so you can call `x = get(chn, [:P, :D])`. This looks like

```julia
(P = (Union{Missing, Float64}[0.349592; 0.671365; … ; 0.319421; 0.298899], Union{Missing, Float64}[0.757884; 0.720212; … ; 0.471339; 0.5381], Union{Missing, Float64}[0.240626; 0.987814; … ; 0.980652; 0.149805]), 
 D = Union{Missing, Float64}[0.648963; 0.0419232; … ; 0.54666; 0.746028])
```

Note that `x.P` is a tuple which has to be indexed by the relevant index, while `x.D` is just a vector.

### DIFF
--- a/src/MCMCChains.jl
+++ b/src/MCMCChains.jl
@@ -18,7 +18,7 @@ using AxisArrays
 const axes = Base.axes
 
 export Chains, getindex, setindex!, chains, setinfo, chainscat
-export describe, set_section, get_all, sections
+export describe, set_section, get_params, sections
 
 # export diagnostics functions
 export discretediag, gelmandiag, gewekediag, heideldiag, rafterydiag
--- a/src/MCMCChains.jl
+++ b/src/MCMCChains.jl
@@ -5,7 +5,7 @@ import StatsBase: autocor, autocov, countmap, counts, describe, predict,
        quantile, sample, sem, summarystats
 import LinearAlgebra: diag
 import Serialization: serialize, deserialize
-import Base: sort, range, names
+import Base: sort, range, names, get
 import Statistics: cor
 
 using RecipesBase
--- a/src/MCMCChains.jl
+++ b/src/MCMCChains.jl
@@ -18,7 +18,7 @@ using AxisArrays
 const axes = Base.axes
 
 export Chains, getindex, setindex!, chains, setinfo, chainscat
-export describe, set_section
+export describe, set_section, get_all, sections
 
 # export diagnostics functions
 export discretediag, gelmandiag, gewekediag, heideldiag, rafterydiag
--- a/src/chains.jl
+++ b/src/chains.jl
@@ -143,14 +143,7 @@ end
 
 
 #################### Indexing ####################
-
-Base.getindex(c::Chains, i1::T) where T<:Union{AbstractUnitRange, StepRange} = c[i1, :, :]
-Base.getindex(c::Chains, i1::Integer) = c[i1:i1, :, :]
-Base.getindex(c::Chains, v::Symbol) = c[[v]]
-Base.getindex(c::Chains, v::String) = Array(c.value[:, [v], :])
-Base.getindex(c::Chains, v::Vector{String}) = Array(c.value[:, v, :])
-
-function Base.getindex(c::Chains, v::Vector{Symbol})
+function _sym2index(c::Chains, v::Vector{Symbol}; sort = true)
     v_str = string.(v)
     idx = indexin(v_str, names(c))
     syms = []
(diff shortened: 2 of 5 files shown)


goedman:
I see your point about shortening get_names to get, but it doesn't give any indication what will be returned. From that point of view something like get_tuples or chain_tuples would be better (but long again). 

The functionality is great! The tuples also give nice access to Parameters.jl facilities such as @unpack.

goedman:
Do we need to make a decision on Symbols vs. Stings when creating Chains? Remember I asked you to look at the problem with my move_parameter_to_segment() method? The problem turns out that I was creating a Chains object using:
```
chn = Chains(val,
  Symbol.(["α", "bp", "_lp"])),
  Dict(
    :parameters => Symbol.(["α", "bp"]),
    :internals => Symbol.(["_lp"])
  )
);
```
and that almost works, but not quite. We can't have a Symbol like :p[1], so maybe the names and Dict arguments in the constructor should enforce Strings? 

I also noted there is a _namedtuple2dict() in chains.jl and utils. There is also a package NamedTupleTools.jl that provides this functionality. Using this package of course generates another dependency. Just fyi.

cpfiffer:
> I see your point about shortening get_names to get, but it doesn't give any indication what will be returned. From that point of view something like get_tuples or chain_tuples would be better (but long again).

I will change it back to `get_params` then, because that also tells people what the arguments should be.

>We can't have a Symbol like :p[1], so maybe the names and Dict arguments in the constructor should enforce Strings? Or :p[1] needs to be specified as Symbol("p[1]").

That occurred to me as well, but it seems to have slipped my mind somewhere between realizing it and putting the pull request up. I'll set everything up to enforce strings for parameter names. Will it be an issue if I do so?

> I also noted there is a _namedtuple2dict() in chains.jl and utils. There is also a package NamedTupleTools.jl that provides this functionality. Using this package of course generates another dependency. Just fyi.

I hadn't noticed! I think we should consider adding that dependency if the `NamedTuple` stuff going on here gets too much more complex, but for right now I think the internal stuff I have here isn't too likely to break.

goedman:
I would be fine with just Strings.

By the way, set_section() works great!

goedman:
After thinking a bit more about it, a few more ideas:

1. Looking how get() is used on a Dict, maybe get() is actually fine.
2. You think it would be useful to have e.g. ```get(chn, :parameters)```?
3. Maybe get_params(chn) can return a flattened list of all parameters?

cpfiffer:
> You think it would be useful to have e.g. `get(chn, :parameters)`?

It might cause some problems in cases where someone has some parameter they've decided to call `"parameters"`, which would cause some ambiguity on the function side. I can add a function `get_all` which explicitly requires a section name.

> Maybe get_params(chn) can return a flattened list of all parameters?

Yeah, absolutely. Good idea.

goedman:
```get_all(chn, :parameters)``` is fine. Or requiring a named argument ```get(chn, section=:pooled)``` is consistent with other calls where we refer to sections.

The ```get_params(chn)``` call, together with ```chn.value.data``` makes it very easy to create DataFrames or NamedArrays or whatever. This way MCMCChains does not need to be dependent on any of those packages (as I suggested a bit earlier).

goedman:
Over the next couple of days I'm going to apply the new MCMCChains features to all of the StatisticalRethinkingJulia packages. Maybe something else pops up, but I think what you have right now is very powerful and elegant!

cpfiffer:
Alrighty, here's a summary of changes:

* `Strings` are now enforced as parameter names. Symbols no longer work.
* Added `get_params`, which packages all the parameters in a chain into a `NamedTuple`. Used like `get_params(chn)`.
* Added an overload on `get` accepting a keyword `section`, which can be a vector or a single symbol. Works like `get(chn; section = :pooled)` or `get(chn; section = [:parameters, :pooled])`. 
* All variants of `get` and `get_params` accept a keyword argument `flatten`, which can be `true` or `false`. If `flatten=true`, the tuple returned will have a flat structure. This is handy if you want to go straight to a dataframe, because `flatten=false` (the default) will clump variables with the same names together (as with `"P[1]", "P[2]", "P[3]"`).

Issue #66, opened 2019-03-13
Sampling capability for posterior predictive distributions


fisherc2:
Currently, MCMCChains does not have built in functions for sampling posterior values from chain objects. As discussed on [discourse](https://discourse.julialang.org/t/posterior-predictive-distributions-in-turing/21781/5), having a sampling function or framework could be useful for generating posterior predictive distributions.

cpfiffer:
Yeah, I think that'd get us a long way. I made a couple changes, like using `range(chn)` instead of `1:size(chn, 1)`, since we allow for chains to have custom ranges. I also changed `Chains(chn.value[indxs, :, :], names(chn), chn.name_map)` to just `return chn[indxs, :, :]`.

```julia
import StatsBase: sample
using Random

# sample([rng], a, [wv::AbstractWeights], n::Integer; replace=true, ordered=false)
function sample(rng::AbstractRNG, chn::MCMCChains.AbstractChains, n;
    replace=true, ordered=false) 
  indxs = sample(rng,range(chn), n,
    replace=replace, ordered=ordered)
  return chn[indxs, :, :]
end

function sample(chn::MCMCChains.AbstractChains, n;
    replace=true, ordered=false) 
  indxs = sample(Random.GLOBAL_RNG, range(chn), n,
    replace=replace, ordered=ordered)
  return chn[indxs, :, :]
end
```

Should this return an array or something instead of a `Chains` object? I'm never sure when it is we should return numeric values vs. a chain.

goedman:
Hmm, thought I started off with your last improvement. But this is great.

My suggestion is to return a chain mainly because of the next step, weighted sampling for posterior sampling, where we might want to work with 10000 samples vs. maybe just 1000 mcmc draws. 

The other reason is related to my response to the other issue (Expand Utility Function), I suggest for now we stick to a rule like turning a chain into numerical values is the last step. We export at that point `out of MCMCChains`. This also preserves consistency and all nice plotting and selection capabilities.

I'll add the weight factors in a similar fashion today.

goedman:
Cameron, I'm using this test program:
```
using StatisticalRethinking
import StatsBase: sample

ProjDir = @__DIR__
cd(ProjDir)

include("../test/samplechain.jl")

describe(chn)

df = to_df(chn)

df_sample = sample(df, 5)
display(df_sample)
println()

chn_sample = sample(chn, 5)
display(chn_sample)
println()

s = kde(df[:sigma])
[...]

PR #68, opened 2019-03-18
Sampling api

goedman:
This PR contains a proposed (simplified) sampling API as defined in StatsBase on top of MCMCChains.Chains objects as requested in #66.

It supports random sampling of draws in chains and sampling based on a Weights vector. This PR overloads StatsBase.sample() for that purpose and uses KernelDensity.kde() to compute a support range and associated densities. It is certainly possible to drop the KernelDensity dependency (although implicitly it is there through the density() plot).

### DIFF
--- a/src/sampling.jl
+++ b/src/sampling.jl
@@ -1,0 +1,50 @@
+import StatsBase: sample
+using Random, KernelDensity
+
+function sample(rng::AbstractRNG, chn::MCMCChains.AbstractChains, n;
+    replace=true, ordered=false) 
+  indxs = sample(rng, 
+    range(chn),
+    n,
+    replace=replace, ordered=ordered)
+  return Chains(chn.value[indxs, :, :],
+    names(chn), 
+    chn.name_map,
+    info = chn.info)
+end
+
+function sample(chn::MCMCChains.AbstractChains, n;
+    replace=true, ordered=false)
+  indxs = sample(Random.GLOBAL_RNG,
+    range(chn),
+    n,
+    replace=replace, ordered=ordered)
+  return Chains(chn.value[indxs, :, :],
+    names(chn),
+    chn.name_map,
+    info = chn.info)
+end
+
+function sample(rng::AbstractRNG, chn::MCMCChains.AbstractChains,
+    wv::AbstractWeights, n) 
+  indxs = sample(rng, 
+    range(chn), 
+    wv, 
+    n)
+  return Chains(chn.value[indxs, :, :],
+    names(chn),
+    chn.name_map,
+    info = chn.info)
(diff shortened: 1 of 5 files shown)


cpfiffer:
I'll merge this over in a while unless @trappmartin has any thoughts to throw in.

trappmartin:
Great, I will have a look at it today.

src/sampling.jl
```diff
@@ -0,0 +1,50 @@
+import StatsBase: sample
+using Random, KernelDensity
+
+function sample(rng::AbstractRNG, chn::MCMCChains.AbstractChains, n;
+    replace=true, ordered=false)
```
yebai: Some comments on the meaning of `replace` and `ordered` might be helpful here.

goedman: True!

In this particular instance, ```sample()```,  StatsBase will step in:
```
? sample
```
will return:
```
help?> sample
search: sample sampler sample! samplepair Sampleable wsample wsample! nsamples

  sample([rng], a, [wv::AbstractWeights], n::Integer; replace=true, ordered=false)

  Select a random, optionally weighted sample of size n from an array a using a
  polyalgorithm. Sampling probabilities are proportional to the weights given in
  wv, if provided. replace dictates whether sampling is performed with replacement
  and order dictates whether an ordered sample, also called a sequential sample,
  should be taken.


  Optionally specify a random number generator rng as the first argument (defaults
  to Random.GLOBAL_RNG).
```

but in general, MCMCChains has enough how-to-use-it and features that it warrants it's own docs.

test/sampling_tests.jl
```diff
@@ -0,0 +1,23 @@
+using Turing, MCMCChains, KernelDensity, StatsBase, Test, Statistics
+import StatsBase: sample
```
trappmartin: Why is this necessary?

test/sampling_tests.jl
```diff
@@ -0,0 +1,23 @@
...
+  sampler = HMC(500, 0.01, 5)
+  chn = sample(model, sampler, save_state=true);
+
+  chn_sample = sample(chn, 5)
+  @test range(chn_sample) == 1:1:5
+
+  c = kde(reshape(convert(Array{Float64}, chn[:s].value), 500))
+  chn_weighted_sample = sample(c.x, Weights(c.density), 100000)
+
+  @test 4.9 < mean(reshape(convert(Array{Float64}, chn[:s].value), 500)) < 5.1
```
trappmartin: Maybe replace with `@test mean(convert(Array{Float64}, chn[:s].value)) ≈ 5.0 atol=0.01`

PR #73, opened 2019-03-26
Added export constructors (I used the wrong local branch name).

goedman:
Added Array and DataFrame constructors and associated tests.

These constructors are intended as being used at the back-end of mcmc workflows, thus after checking for valid samples has been completed. Both the Array and DataFrame constructors are overloaded.

Simple help is available using ?Array and ?DataFrame once MCMCChains is loaded.

### DIFF
--- a/src/constructors.jl
+++ b/src/constructors.jl
@@ -1,0 +1,205 @@
+function sort_sections(chn::MCMCChains.AbstractChains)
+  smap = keys(chn.name_map)
+  section_list = Vector{Symbol}(undef, length(smap))
+  indx = 1
+  if :parameters in smap
+    section_list[1] = :parameters
+    indx += 1
+  end
+  if :internals in smap
+    section_list[end] = :internals
+  end
+  for par in smap
+    if !(par in [:parameters, :internals])
+      section_list[indx] = par
+      indx += 1
+    end
+  end
+  section_list
+end
+
+"""
+
+# Array
+
+Array constructor from an MCMCChains.Chains object. Returns 3 dimensionsal
+array or and Array of 2 dimensional Arrays. If only a single parameter if selected for
+inclusion, a dimension is dropped in both cases, as is e.g. required by cde(), etc.
+
+### Method
+```julia
+  Array(
+    chn::MCMCChains.AbstractChains, 
+    sections::Vector{Symbo);
+    append_chains::Bool,
+    remove_missing_union::Bool
+  )
+```
+
+### Required arguments
+```julia
+* `chn` : Chains object to convert to an Array
+* `sections = Symbol[]` : Sections form the Chains object to be included
+```
+
+### Optional arguments
+```julia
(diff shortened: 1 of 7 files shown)


trappmartin:
Looks good. 

However, I'm not 100% convinced why it makes sense to implement those constructors instead of simply providing `convert` function, e.g. `convert(::Type{Array}, chn::Chains) = chn.value`?
Maybe provide both?

src/constructors.jl
```diff
@@ -0,0 +1,205 @@
...
+
+Array constructor from an MCMCChains.Chains object. Returns 3 dimensionsal
+array or and Array of 2 dimensional Arrays. If only a single parameter if selected for
+inclusion, a dimension is dropped in both cases, as is e.g. required by cde(), etc.
+
+### Method
+```julia
+  Array(
+    chn::MCMCChains.AbstractChains, 
+    sections::Vector{Symbo);
```
cpfiffer:
```suggestion
    sections::Vector{Symbol};
```

src/constructors.jl
```diff
@@ -0,0 +1,205 @@
...
+  end
+  section_list
+end
+
+"""
+
+# Array
+
+Array constructor from an MCMCChains.Chains object. Returns 3 dimensionsal
+array or and Array of 2 dimensional Arrays. If only a single parameter if selected for
```
cpfiffer:
```suggestion
array or an Array of 2 dimensional Arrays. If only a single parameter is selected for
```

src/constructors.jl
```diff
@@ -0,0 +1,205 @@
...
+    sections::Vector{Symbo);
+    append_chains::Bool,
+    remove_missing_union::Bool
+  )
+```
+
+### Required arguments
+```julia
+* `chn` : Chains object to convert to an Array
+* `sections = Symbol[]` : Sections form the Chains object to be included
```
cpfiffer:
```suggestion
* `sections = Symbol[]` : Sections from the Chains object to be included
```

cpfiffer:
As to what @trappmartin said, I added a `convert(::Type{Array}, chn::Chains)` function just to have around. It's super basic in case someone just needs a smash-and-grab array. Rob's `Array` functions is much better I think for handling all the sometimes weird stuff MCMCChains has.

goedman:
I might have jump the gun by merging the export chains additions. My apologies, I thought I'd seen a generic `approved` tick. 

For quite a while I've used the convert() option but often felt baffled by not having control over the exact sequence in which variables are inserted into the array or dataframe. This is particularly bothersome for arrays that don't hold the names. The other considerations are appending chains, filtering out the Missing union and above all an option to work with one or more specific sections, e.g. population wide parameters and pooled parameters.

Today I've added some short sections to the README (sampling and exports) and then we could maybe consider a new release. I just did a PR with these changes included.

I'm looking into one other component which is a DataFrame based chain summary, but that is still early on.

Issue #71, opened 2019-03-21
Make chain summaries easier to extract and store


itsdfish:
Hi-

I am opening this issue based on a discussion in [discourse](https://discourse.julialang.org/t/extracting-summary-info-from-multiple-mcmcchains-programmatically/22027/5).  The current problem is that summary information is difficult to extract from the chain programmatically in large scale simulations. The current solution is to use StatsBase and extract from the chain's nested structure:

```
using StatsBase

function getSummary(ch)
    chain = ch[1001:end,:,:]
    return summarystats(chain).summaries[1].value
end
```

So the question is whether there is a simpler way for the user to work with summary data.

goedman:
Cameron, I have been wondering about [Tables.jl](https://github.com/JuliaData/Tables.jl). Of course yet another dependency. It provides an Array with named rows and columns. The parameters could be the rows, the statistics (mean, sd, ...) the columns. I've never worked with this package and would like to check if we can enable stuff like get(table, [:alpha, :beta], [:mean, sd]).

cpfiffer:
Tables does seem to be low-level enough and a fairly small dependency, so it might be advantageous to use. Plus, having easily modifiable columns would allow use to get rid of `Showoff` which I'm not a big fan of.

goedman:
Yes, although I'm not sure with Tables we get a lot of benefits in terms of printout/format support. It would be nice to get that for free. The other option is just DataFrames with a 1st column of Symbols. How would you feel about that? 

In that case, like the Array proposal, we can also support exporting chains as a DataFrame.

In fact, I'll make a pull request that contains both (Arrays and DataFrames) to play around with. I have added one more aspect to the Array constructor. If you export a single parameter the output format is collapsed into a vector, e.g. kde requires that. In a sense Arrays and DataFrames are nicely complementary.

mkborregaard:
DataFrames depends on Tables, so. IMHO DataFrames is a good implementation for object-internal storage, but for input arguments it's better to use the Tables interface than to require users to have their stuff in DataFrames.

goedman:
To make sure I understand your comment correctly Michael, the proposed usage is that after running an mcmc simulation, MCMCChains creates this table with results once and stores it, e.g. show() would simply display the table. In some cases folks would like to be able to programmatically read a few fields, e.g. for tests in runtests.jl or, as Chris might use it, to compare many mcmc runs programmatically. I see it as output only, not for user input. 

[...]

PR #79, opened 2019-04-09
Use DataFrames for ChainSummaries

cpfiffer:
This PR switches MCMCChains off the strange I/O driven `ChainSummary` and `ChainSummaries` structs, and uses a much simpler `ChainDataFrame` struct that brings a lot of the versatility of `DataFrames` for printing, organization, and indexing. It's now a lot easier to get values out of `ChainSummary` structs (see #71).

Currently, the bulk of the work is done. There's still a couple of remaining steps, namely tidying and making sure all the tests are doing what they're supposed to:

- [x] Remove `ChainSummary` and `ChainSummaries` entirely
- [x] Make all the tests pass

There is a new function called `summarize` which maps a function parameter-wise and returns a dataframe. Hopefully this'll make it quicker to get some non-traditional summary stats, and it's used a lot on the backend. Example:

```julia
using MCMCChains, Statistics

chn = Chains(rand(500, 3, 2))
x = summarize(chn, mean, var, std)
```
```
│ Row │ parameters │ mean     │ var       │ std      │
│     │ Symbol     │ Float64  │ Float64   │ Float64  │
├─────┼────────────┼──────────┼───────────┼──────────┤
│ 1   │ Param1     │ 0.49674  │ 0.0877798 │ 0.296277 │
│ 2   │ Param2     │ 0.492215 │ 0.0807893 │ 0.284235 │
│ 3   │ Param3     │ 0.49734  │ 0.0850419 │ 0.291619 │
```

The indexing on these is overloaded to put a higher priority on rows:

```julia
x[:mean] # Select the mean column
x[:Param1, :mean] # select the mean for Param1
x[:Param1, :] # Select the Param1 row
```

The `show` function for `Chains` now uses this as well:

```julia
julia> chn
Object of type Chains, with data of type 500×3×2 Array{Float64,3}

Iterations        = 1:500
Thinning interval = 1
Chains            = 1, 2
Samples per chain = 500
parameters        = Param1, Param2, Param3

Summary Statistics

│ Row │ parameters │ mean     │ std      │ naive_se   │ mcse       │ ess     │
│     │ Symbol     │ Float64  │ Float64  │ Float64    │ Float64    │ Float64 │
├─────┼────────────┼──────────┼──────────┼────────────┼────────────┼─────────┤
│ 1   │ Param1     │ 0.49674  │ 0.296277 │ 0.00936909 │ 0.0109575  │ 731.097 │
│ 2   │ Param2     │ 0.492215 │ 0.284235 │ 0.00898829 │ 0.00939995 │ 914.329 │
│ 3   │ Param3     │ 0.49734  │ 0.291619 │ 0.00922182 │ 0.00675694 │ 1000.0  │
```
All the stats functions include keyword arguments for `sections` and `append_chains`. If `append_chains=true`, then all the chains are smushed together into one chain. Otherwise, you'll get a vector of `ChainDataFrames`, one for each chain. `sections` just allows you to subset your chain by sections.


cpfiffer:
It looks like we've been calculating ESS using 

```julia
min((std(x) / df_mcse(x))^2, length(x))
```

which seems really strange to me. Shouldn't we use something more like what [Stan](https://mc-stan.org/docs/2_18/reference-manual/effective-sample-size-section.html) (and lots more) use?

If nobody is opposed, I'm going to update our ESS calculator to use the autocorrelation-derived variant.

goedman:
Hi Cameron, I've been looking at https://arxiv.org/abs/1903.08008 . Have you seen that update?

cpfiffer:
I hadn't, no. Funnily enough I was sitting down with the "Bayesian Data Analysis" book by Gelman et al. (2014) which implements this ESS.

Maybe we should open a separate issue to address some of the points brought up in the article you mentioned.

goedman:
Good idea. 

I first want to look at WAIC and that modelchain stuff in MCMCChains. 

Also of interest on LOO etc might be, a bit older ( http://www.jmlr.org/papers/volume11/watanabe10a/watanabe10a.pdf ). Maybe we should make that a separate issue as well.

cpfiffer:
I've added an `ess` function which liberally lifts from the [Stan implementation](https://github.com/stan-dev/stan/blob/11d5b7b8d3edbc2dfd4883859e741ec9b156aa5e/src/stan/analyze/mcmc/compute_effective_sample_size.hpp#L25). It'll show up now whenever `describe`, `summarystats`, or `show(chn)` is called.

cpfiffer:
Looks like our tests are passing. This is now ready for a formal review.

goedman:
I think this looks great Cameron!

trappmartin:
I'm happy to review the PR by tomorrow.

goedman:
Cameron, just as a thought, would it be useful to have `describe()` return the summary data frame?
Working through the CmdStan examples I think I'll end up calling both most of the time.

cpfiffer:
I actually wasn't sure what to do about that, so I'm happy to take suggestions there. My thinking as to why it doesn't actually return anything is that I had thought it was intended as primarily an I/O function to show you information about the chain --- not to return the dataframes. I had it set to do both but if you ran `describe` from the REPL it would print the value out twice: one for the `show` calls in the function, and one for the return value.

Perhaps a better option would be to have `describe(chn)` only return the dataframes (with no printing or internals `show` calls), and then set `show(chn)` to call `describe` in the back end. Does that sounds slightly more reasonable?

src/discretediag.jl
```diff
@@ -416,11 +416,60 @@ end
 #  return [p1, p2]
 #end
 
+# function discretediag(chn::AbstractChains; frac::Real=0.3,
```
trappmartin: Cant this be removed?

src/discretediag.jl
```diff
@@ -440,15 +489,18 @@ function discretediag(chn::AbstractChains; frac::Real=0.3,
     V, vals = discretediag_sub(c, frac, method, nsim,
     size(c.value,1), size(c.value,1))[1:2]
 
-    #println(V)
-    #println(vals)
+    num_chains = length(chains(c))
+    sep_vals = [vals[(3 * (k - 1) + 1):(3 * (k - 1) + 3), :]'
```
trappmartin: Maybe add a comment what is happening here.

trappmartin:
Look good to me. I would not add any further functionalities as PR is already quite large. Better merge the PR and open a new one for the `describe` function.

goedman:
Yes, running it in the REPL would require `s = describe(chns);` I guess, with the `;`. I could live with that. Part of my suggestion is to prevent, for very big models, generating the data frame twice. But in general the time the sampling takes dominates the post processing anyway.

cpfiffer:
Merging for now, since I've only modified comments and previous tests were passing. Thanks for the review, guys!

Issue #81, opened 2019-04-10
Make density and mixeddensity unique


itsdfish:
Hi all-

As far as I can tell,  `density` and `mixeddensity` generate the same plot. It might be useful to change the function of `mixeddensity` to be a mixture of each chain. In other words, all of the samples would be pooled into one array and plotted. I think plotting individual chains is helpful for diagnostics and pooling is useful for generating uncluttered plots for publication.

mkborregaard:
`mixeddensity` allows showing a histogram when chains are discrete.

fisherc2:
Thanks for the info. In that case, I wonder if something like` :pooleddensity` would be useful. If not, I think we can close this issue.

cpfiffer:
Currently I believe what you're talking about exists. Try `plot(chn, colordim = :parameter)` to plot pooled densities. Does this do what you've asking about?

goedman:
Doesn't just `plot(chn)` provide that?

cpfiffer:
`plot(chn)` gives you a plot for each parameter with all chains plotting for the same parameter, while `plot(chn, colordim=:parameter)` gives you one plot per chain and throws all the parameters on that one plot. The README has this example:

Summarize parameters | Summarize chains
:-------------------------:|:-------------------------:
`plot(chn; colordim = :chain)` | `plot(chn; colordim = :parameter)`
![p1](https://user-images.githubusercontent.com/7974003/45822242-f0009180-bce2-11e8-8fa0-a97c8732400f.png)  |  ![p2](https://user-images.githubusercontent.com/7974003/45822249-f131be80-bce2-11e8-8dd3-42db7d58abd9.png)

fisherc2:
Just to clarify my initial post, I would like to do something that is similar to `density(chn[:Param1].value[:])` while maintaining the labels and settings that are used in the builtin plot functions. This would pool the samples across chains for a given parameter. As far as I can tell, none of the options discussed so far have this functionality.

cpfiffer:
Oh, try indexing the chain first if you only want one parameter. `plot(chn[:Param1], colordim=:parameter)`. I think it's possible that `plot(chn, :Param1, colordim=:parameter)` might also work, but I'm not at a good machine to test that out right now.

fisherc2:
Unfortunately, that does not work. It generates a subplot for each chain. Here is a simple example based on the Turing docs:


```
using Distributed,StatsPlots
addprocs(4)
[...]

PR #84, opened 2019-04-14
Allow for pooled density plots

cpfiffer:
Addresses #81.

`pooleddensity(chn)` (a new function) will now generate a histogram (if discrete) or a density plot (if continuous) by pooling parameters together. You can replicate this behavior with the other plotting function by calling them with the keyword `append_chains = true`.

@itsdfish does this seem reasonable to you?

Example:

```julia
using MCMCChains, StatsPlots, Distributions
d = Normal()
chn = Chains(rand(d, 5000, 2, 10))

# Regular plotting.
plot(chn)
```

```julia
# Append the chains together.
plot(chn, append_chains=true)
```

```julia
# Used pooleddensity
pooleddensity(chn)
```


itsdfish:
It looks great to me. Thanks!

PR #98, opened 2019-04-30
ESS Fix

### LINKED ISSUE #90: ESS appears to be incorrect
I noticed two problems with ESS. First, it appears to always be ESS=Nsamples. Second, when I  excluded the warmup samples, ESS=0. Here is a simple example:

```
using Turing, StatsPlots, Random

Random.seed!(14)
data = rand(Normal(0,1), 30)

@model model(y) = begin
    μ ~ Normal(0, 10)
    N = length(y)
    for n in 1:N
        y[n] ~ Normal(μ,1)
    end
end;

# Settings of the NUTS sampler.
Nsamples = 2000
#Target acceptance rate
δ = .8
#The number of adaption or warmup samples
Nadapt = 1000
#Collects sampler configuration options
specs = NUTS(Nsamples,Nadapt,δ)

# Start sampling.
chain = sample(model(data),specs);

plot(chain,seriestype=(:autocorplot))

#Produces ESS = 0
subset = chain[(Nadapt+1):end,:,:]
```
Autocorrelation is good but it ESS should be less than Nsamples.

cpfiffer:
Fixes #90 by actually doing the math (mostly) right. ESS now also reports Rhat, as below:

```julia
ess(chn)
```
```julia
ESS

│ Row │ parameters │ ess     │ r_hat    │
│     │ Symbol     │ Any     │ Any      │
├─────┼────────────┼─────────┼──────────┤
│ 1   │ theta      │ 1558.33 │ 0.999004 │
```

Rhat will also show up in the summary stats block.


src/stats.jl
```diff
@@ -208,29 +208,29 @@ function ess(chn::AbstractChains;
...
-            eachchain[j] = mean(draw[j])
-            s[j] = (1/(n-1)) * sum((draw[j] .- eachchain[j]).^2)
-        end
+        eachchain = [mean(draw[j]) for j in 1:m]
+        s = [sum((draw[j] .- eachchain[j]).^2) / (n-1) for j in 1:m]
         B[i] = (n / (m - 1)) * sum((eachchain .- allchain).^2)
-        W[i] = (1/m) * sum(s.^2)
-        varhat[i] = (n-1)/n * W[i] + (1/n) * B[i]
+        W[i] = sum(s) / m
+        varhat[i] = (n-1)/n * W[i] + (1/n) * (B[i] / n)
```
benelsen: I think one of the division of the between-chain variance by the chain length is superfluous here

cpfiffer: Ah, good catch. Thanks!

goedman:
Just out of interest, in addition to the ess plot, a plot for r_hat. Below both plots.


[ess__estimates_plot.pdf](https://github.com/TuringLang/MCMCChains.jl/files/3135350/ess__estimates_plot.pdf)
[rhat__estimates_plot.pdf](https://github.com/TuringLang/MCMCChains.jl/files/3135351/rhat__estimates_plot.pdf)

itsdfish:
Rob-

I have a quick question about the r-hat plot. If I understand correctly, it seems like Turing produces larger and more variable r-hat than CmdStan for the same model, data and r-hat function. However, it appears that the delta parameter for Turing is .95 and .80 for CmdStan. Is this causing the difference and is that intentional?

goedman:
Yes, both the ess and r-hat show much greater variability, which surprised me as well. Early on I did use runs with the accept target at .65 and .85, and saw no difference for ess. Will run the test with .80, just to make sure.

itsdfish:
Equating the target acceptance rate seems like a good idea. If the problem persists, it suggests there might be a problem with Turing somewhere.

goedman:
These are the plots with the accept target at .80:

[ess__estimates_plot.pdf](https://github.com/TuringLang/MCMCChains.jl/files/3136344/ess__estimates_plot.pdf)
[rhat__estimates_plot.pdf](https://github.com/TuringLang/MCMCChains.jl/files/3136345/rhat__estimates_plot.pdf)

itsdfish:
I've had a suspicion for a while that Turing is not producing the right results, especially for autocorrelation/ESS. It might be worth filing an issue unless you can think of some other explanation for the large divergence.

goedman:
And with delta = 0.65:

[ess__estimates_plot.pdf](https://github.com/TuringLang/MCMCChains.jl/files/3136374/ess__estimates_plot.pdf)
[rhat__estimates_plot.pdf](https://github.com/TuringLang/MCMCChains.jl/files/3136377/rhat__estimates_plot.pdf)

goedman:
Will do that when back in DC an check on Slack first. Of course I’m the world’s most limited Turing user, just Nuts.

itsdfish:
That sounds good. Let me know if you want me to file an issue to help out.  I just wanted to make sure you couldn't think of any other explanations for the behavior.

PR #102, opened 2019-05-17
Update chains.jl

dalejordan:
get isn't passing on the flatten parameter to lower level get.


cpfiffer:
Good catch, thank you for sending a PR! I'll merge this over when the tests pass.

cpfiffer:
Sorry about the delay on this one, there were some breaking tests unrelated to this PR that had to be fixed first.

PR #103, opened 2019-05-18
Remove Turing from dataframe/summarize tests

cpfiffer:
Builds on Turing master and AdvancedHMC master are failing. I think it's a good idea to remove Turing from the summarize and DataFrame constructor tests since Turing's internals has had and will maybe continue to have some changes.

### DIFF
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -1,0 +1,479 @@
+# This file is machine-generated - editing it directly is not advised
+
+[[AbstractFFTs]]
+deps = ["LinearAlgebra"]
+git-tree-sha1 = "380e36c66edfa099cd90116b24c1ce8cafccac40"
+uuid = "621f4979-c628-5d54-868e-fcf4e3e8185c"
+version = "0.4.1"
+
+[[Arpack]]
+deps = ["BinaryProvider", "Libdl", "LinearAlgebra"]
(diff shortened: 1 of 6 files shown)


goedman:
Hi Cameron, that’s fine. Is that also the case for DynamicNUTS variant as far as you know?

cpfiffer:
As far as I'm aware, DynamicNUTS is safe. I will send you a message if I notice that changing, though.

goedman:
Thanks, Chris and I are doing a lot of benchmarking on mcmc and if possible we do like to continue to keep DynamicNUTS in the loop.

cpfiffer:
Yeah, there's no plans to drop it. It's a good implementation and it plays well with our internals, so we're happy to keep it along.

cpfiffer:
Most of what I was referring to with changes have to do with Turing's internal HMC stuff being moved over to AdvancedHMC, which is a lot cleaner and faster than the older stuff. It doesn't use the same `:internals` parameters as Turing's original HMC stuff. Don't worry -- we're not going to run around and break everything.

goedman:
You've been pretty solid in that respect (and others), so I'm not worried. It is interesting to see how much is changing in Julia's mcmc world! Good stuff!

PR #127, opened 2019-08-28
Clean up and DIC

trappmartin:
This PR does some clean up in the codebase and adds a correct implementation of the DIC & a small unit test.


trappmartin:
Sounds like a good idea. We can probably easily adapt the code once we have this. The main idea of this PR is to have a first implementation of DIC so that people can contribute implementations of AIC, WAIC and so on.

Issue #126, opened 2019-08-18
Issues with DataFrames/


goedman:
Hi Cameron,

I noticed recently the following warning:
```
┌ Warning: `getindex(df::DataFrame, col_ind::ColumnIndex)` is deprecated, use `df[!, col_ind]` instead.
```
It happens when I select from a ChainDataFrame:
```
cdf[:theta, :mean]
```
I think it was introduced by the latest DataFrames update that expect the `df[!, ...]` syntax.

Initial testing with Julia 1.4, as Julia 1.3.0-rc1 gave problems, might show another issue with both DataFrames and MCMCChains exporting `describe`. But I need to further check that.

goedman:
Hi Cameron, I tested with 2 updates in summarize.jl:
lines 71 & 92: `.df[:,`   -> .`df[!,`

Not sure if those are all cases, but it stops the warnings when I test Stan.jl.

cpfiffer:
I'll get a PR in to fix this everywhere, thanks!

yebai:
Is this fixed now?

goedman:
Yes, I believe it is fixed!

PR #131, opened 2019-09-03
Fixes DataFrames indexing

cpfiffer:
Addresses #126 by fixing deprecated behavior in DataFrames.jl. This PR also fixes a minor bug where mutable name maps like a `Dict` were being mutated by the `Chains` constructor.

### DIFF
--- a/src/chains.jl
+++ b/src/chains.jl
@@ -54,18 +54,20 @@ end
 function Chains(
         val::AbstractArray{A,3},
         parameter_names::Vector{String} = map(i->"Param$i", 1:size(val, 2)),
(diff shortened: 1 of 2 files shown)


goedman:
Thank you! About to take off for Europe, but will test over the next couple of days. A bit surprised you can still use df[:, ...], thought the new syntax is now df[!, ...].

cpfiffer:
Yeah, `df[:, :something]` is for non-mutating retrieval, whereas `df[!, :something]` is sort of explicitly saying "I'm going to mutate all the rows now".

PR #135, opened 2019-10-22
Add set_names to README

itsdfish:
I had trouble finding `set_names` and noticed it was not in the documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,21 @@ chn = Chains(val, ["a", "b", "c", "d", "e"])
 
 By default, parameters will be given the name `Parami`, where `i` is the parameter number.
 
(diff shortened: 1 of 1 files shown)


itsdfish:
I'm not sure exactly what happened here. It looks like it could not install Julia. I don't think the changes to the docs would have caused the problem.

goedman:
HI Chris, I think this is a broader Julia on Travis issue. All Travis installs for Julia seem to fail.

itsdfish:
Yikes. That's no good. Thanks for the info.

cpfiffer:
Since there's no functional change, I'll just merge this over. Thanks, Chris!

PR #136, opened 2019-10-28
Remove Plots and fix tests

andreasnoack:
Plots was in the project file but wasn't used. It's a fairly heady dependency so would be avoid since it isn't needed. I also fixed the tests for the changes to Turing and put a version requirement on Turing.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -5,16 +5,12 @@ license = "MIT"
 desc = " Chain types and utility functions for MCMC simulations."
 version = "0.3.14"
 
-[compat]
-julia = "1"
-
 [deps]
 AxisArrays = "39de3d68-74b9-583c-8d2d-e117c070f3a9"
 DataFrames = "a93c6f00-e57d-5684-b7b6-d8193f3e46c0"
 Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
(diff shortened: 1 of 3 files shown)


cpfiffer:
Good catch, thank you!

andreasnoack:
So it looks like all the tests failures aren't actually fixed. See https://travis-ci.org/TuringLang/MCMCChains.jl/jobs/601326701. However, I'm not completely sure what's the right fix. Based on the diffs in Turing.jl I assumed something like
```
diff --git a/test/serialization_tests.jl b/test/serialization_tests.jl
index 2edcd7e..11e8b76 100644
--- a/test/serialization_tests.jl
+++ b/test/serialization_tests.jl
@@ -30,7 +30,7 @@ ProjDir = mktempdir()
         open(f->read(f, String), joinpath(ProjDir, "chn2.txt"))

     # Test whether sampler state made it through serialization/deserialization.
-    chn3 = Turing.Utilities.resume(chn2, 100)
+    chn3 = sample(chn2, 100, resume_from=chn2)
     @test range(chn3) == 1:1:600
 end
```
should work but `chn3` only has 100 samples. Is that expected?

trappmartin:
Resume used to resume from the old chain and return the full result, not only the new samples. I think this is a bug introduced through the interface change.

PR #140, opened 2019-11-08
Don't merge variables that are substrings

sethaxen:
This fixes a bug where multivariate variables that were substrings of another were combined. For example, if a variable `theta[1]` and a variable `eta[1]` were in the `Chains`, then calling `get` with `flatten=false` would give an `eta` with both the values of `eta` and of `theta`.


cpfiffer:
Looks good to me. Test failures are on the Turing side, which I'll address over there. Thanks for your crazy level of energy, @sethaxen.

sethaxen:
Don't thank me, thank the ☕️.  I am but a humble vessel.

PR #144, opened 2019-11-10
Create CompatHelper.yml

cpfiffer:


goedman:
Hi Cameron,

Have you considered to update the last `run` command to"
```
      ...
      - name: Pkg.add("CompatHelper")
        run: julia -e 'using Pkg; Pkg.add("CompatHelper")'
      - name: CompatHelper.main()
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
        run: >-
          julia -e '
          using CompatHelper;
          CompatHelper.main() do;
            run(`julia --project=test -e "import Pkg; Pkg.instantiate(); Pkg.update()"`);
            run(`julia --project=docs -e "import Pkg; Pkg.instantiate(); Pkg.update()"`);
          end
          '
```
If I understand it correctly this would test the package PR on Travis with the updated upper bound(s). I'm not 100% sure about if this is only true if Manifest.toml is stored in the repository.

I've started to install CompatHelper.yml as well to see how it works but haven't seen a PR yet. But in many cases I only added upper bounds on major version numbers (thus willing to take the risk a minor update to a dependency breaks my package).

cpfiffer:
My understanding so far is that CompatHelper opens a PR on a repo, and then the tests should run on those PRs as they would for any other PR. I think in this case it's good to just let CompatHelper only open PRs and then we can make sure tests run accordingly with Travis or Azure or whatever.

PR #145, opened 2019-11-13
Use reduce instead of splatting

devmotion:
For stacking vectors `v` of vectors or matrices, `reduce(hcat, v)` is usually more efficient than `hcat(v...)`, due to the optimized implementation in https://github.com/JuliaLang/julia/blob/89a51fb98485219680c7106e94cf75fd8069f3f8/base/abstractarray.jl#L1375-L1379 (see also the discussion in https://github.com/JuliaLang/julia/issues/21672).

A simple benchmark demonstrates the performance increase with `reduce`:
```julia
using BenchmarkTools

f(x) =  Array(hcat(v...)')
g(x) = copy(reduce(hcat, x)')

u = [rand(128) for i in 1:1000]
@btime f($u)        # 506.439 μs (9 allocations: 1.95 MiB)
@btime g($u)        # 365.840 μs (4 allocations: 1.95 MiB)

v = [rand(1000) for i in 1:128]
@btime f($v)        # 529.852 μs (9 allocations: 1.95 MiB)
@btime g($v)        # 272.787 μs (4 allocations: 1.95 MiB)
```

BTW, the possibly more idiomatic implementation with `mapreduce` is terribly slow:
```julia
h(x) = mapreduce(transpose, vcat, x)

@btime h($u)        # 91.605 ms (3982 allocations: 488.89 MiB)
@btime h($v)        # 11.339 ms (508 allocations: 63.00 MiB)                           
```

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -20,8 +20,6 @@ Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 
 [compat]
-julia = "^1"
-Turing = "^0.7"
 AxisArrays = "^0.3"
 DataFrames = "^0.19"
 Distributions = "^0.21"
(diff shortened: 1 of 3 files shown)


devmotion:
The test errors are the same as on the master branch and due to the reorganization of the `resume` functionality in Turing, it seems.

cpfiffer:
Thanks for the PRs, much appreciated!

PR #160, opened 2019-12-14
Remove chain sorting by default

cpfiffer:
This PR makes it so that chain parameters are no longer sorted by default -- any packages that relied on MCMCChains using natural sort order will need to pass `sorted=true` in their `Chains` call in order to keep current behavior.

As per @goedman's suggestion, I have also incremented the version number to 1.0 since MCMCChains is relatively stable at this point.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -3,7 +3,7 @@ uuid = "c7f686f2-ff18-58e9-bc7b-31028e88f75d"
 keywords = ["markov chain monte carlo", "probablistic programming"]
 license = "MIT"
 desc = " Chain types and utility functions for MCMC simulations."
-version = "0.4.1"
+version = "1.0.0"
 
 [deps]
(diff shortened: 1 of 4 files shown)


haozhangphd:
looks like after #218, there is no longer a way of disabling sorting anymore, as the `sort` kwarg was removed from the Chains constructor.

It would be very valuable to have the option of disabling sorting, because for Bayesian neural networks with millions of parameters, sorting can takes ~30 times longer time than sampling, and this time is entirely wasted.

devmotion:
The data is not sorted in MCMCChains anymore, is it? Sorting has to be performed manually.

devmotion:
If you work with Turing, sorting happens here: https://github.com/TuringLang/Turing.jl/blob/46c05895d7e51028211a856c78b1cca60838c5d4/src/inference/Inference.jl#L359

haozhangphd:
Yes, looks like there no way of disabling sorting in Turing. I think it would be best to have an option for users to disable it.

Issue #171, opened 2020-01-31
Long import times for Chains during testing


itsdfish:
I am opening this issue to document long import times for Chains. In some cases, it can take upwards of an hour locally, causing times outs on [travis](https://travis-ci.com/itsdfish/AdvancedPS.jl/jobs/281460102). 

In the following [MWE](https://github.com/itsdfish/HangTest) with fewer dependencies, the tests complete faster, but are nonetheless very long. 

On my local machine, I have found that calling ` import MCMCChains: Chains` slows down testing significantly, and the slow down is particularly pronounced with Requires 1.0 on the first run. This finding is particularly important for testing on travis because it uses a fresh installation and will timeout in larger packages.  Here are my results: 

```
# First Run 
#                No Chains  Chains
#Requires 0.5:   1.82        14.92
#Requires 1.0:   1.55       129.78

# Second Run
#                No Chains  Chains
#Requires 0.5:   0.72       07.06
#Requires 1.0:   0.71       15.41

```

The general pattern has been reproduced on travis: 

With the import [statement](https://travis-ci.com/itsdfish/HangTest/jobs/282109118): about 1 minute. 
Without the import [statement](https://travis-ci.com/itsdfish/HangTest/jobs/282114861): about 7 minutes.

devmotion:
Have to tried to only load MCMCChains? And to downgrade MCMCChains? Would be interesting to see if one can figure out if it is caused by a specific combination of packages and if a specific commit caused a performance regression.

itsdfish:
Good point. I was able to reproduce that pattern of results with MCMCChains 0.4.1. MCMCChains .3.15 had the same pattern of results but ran at 51 seconds with Requires 1.0, instead of 130ish. I could not run earlier versions without errors.

devmotion:
Could you post the full list of packages (output of `] st --manifest`) that were installed and used in these runs?

itsdfish:
For MCMCChains 3.4:

```
(v1.3) pkg> st --manifest
    Status `~/.julia/environments/v1.3/Manifest.toml`
  [7d9fca2a] Arpack v0.4.0
  [68821587] Arpack_jll v3.5.0+2
  [39de3d68] AxisArrays v0.4.0
  [b99e7846] BinaryProvider v0.5.8
  [9a962f9c] DataAPI v1.1.0
  [864edb3b] DataStructures v0.17.9
  [31c24e10] Distributions v0.22.4
[...]

PR #175, opened 2020-02-13
Clean MCMCChains, remove type piracy, and do not invalidate methods in Base

devmotion:
This PR tries to clean and improve MCMCChains quite drastically. It removes two dependencies (Showoff and KernelDensity) which do not seem to be used anywhere in the package. Moreover, it fixes some severe issues with type piracy and invalidating methods in Base (exactly as Tim Holy suspected in [the discourse thread](https://discourse.julialang.org/t/slowdown-after-package-load/33533/2?u=devmotion), it seems), so hopefully this improves https://github.com/TuringLang/MCMCChains.jl/issues/171 as well (there is a similar issue in AbstractMCMC as well IMO).


goedman:
Nice job David! Much cleaner. Haven't checked all my examples yet but until now no surprises!

PR #179, opened 2020-02-17
Fix concatenation on Julia 1.2

devmotion:
This PR fixes the issues on previous Julia versions that we've seen in https://github.com/TuringLang/Turing.jl/pull/1116. Apart from minor fixes I did not touch the actual concatenation algorithm; I'm wondering if most of the implementation could be removed by using `cat(...)` on the `AxisArray`s and avoiding rebuilding a completely new axis array from scratch.


test/concatenation_tests.jl
```diff
@@ -0,0 +1,125 @@
...
+    @test chn2c.name_map == chn2.name_map
+    @test chn2c.info == chn2.info
+
+    # concatenate a different chain
+    chn3 = vcat(chn, chn1)
+    @test chn3.value.data == vcat(chn.value.data, chn1.value.data)
+    @test size(chn3) == (15, 5, 2)
+    @test names(chn3) == names(chn)
+    @test range(chn3) == 1:15
+    @test chn3.name_map == (internal = ["a", "b"], parameters = ["c", "e", "d"])
```
devmotion: Is this the intended behaviour? With the current implementation, in every concatenation only the `name_map` of the final argument is kept (even in case of `hcat`) if the chains use the same sections.

cpfiffer: I believe it was intended, but I don't think it's a particularly good idea. I suppose the ideal case would be to should construct the set addition version of the name map.

devmotion: But that only makes sense for `hcat`, doesn't it? There the different arguments are supposed to have unique names, so a set addition seems reasonable. However, for `vcat` and `chainscat` maybe we should just demand that all maps are equal?

devmotion:
I have no clue why suddenly the tests time out on Julia 1.2 and 1.3, it looked fine on my computer. The main issue seems to be the plotting tests, since they take so much time that the jobs are canceled, but the plotting recipes are not modified by this PR at all?!

devmotion:
The latest changes (and commenting out the plotting tests :stuck_out_tongue:) seems to fix the time out issues. Only the diagnostics (and/or the compilation?) on Julia 1.3 takes quite some time.

devmotion:
As soon as I reenable the plotting tests it takes forever on Julia 1.2 and 1.3. The other versions seem to be fine.

test/concatenation_tests.jl
cpfiffer: Sure, that's better.

test/runtests.jl
```diff
@@ -1,36 +1,14 @@
 using Test
 
-@testset "MCMCChains" begin
```
cpfiffer: This is very difficult to read -- can we maintain the original formatting?

cpfiffer:
I am very confused as to what's happening.

test/runtests.jl
devmotion: The problem with having one MCMCChains testset is that there is no information (and update while running) of the nested testsets. Moreover, all tests will be run even if already the first testset errors. Additionally, even if everything runs reasonably fast, we might have to introduce print statements to just to prevent Travis from cancelling the job.

devmotion:
Yes, I don't see why any of the changes would cause problems.

test/runtests.jl
cpfiffer: We can keep the separate test sets, but they should be nested within an `MCMCChains` set, and should all be on separate lines so they are easier to read. Something like

```julia
@testset "MCMCChains" begin
    @testset "Diagnostics" begin 
        include("diagnostic_tests.jl") 
    end

    . . .
end
```

We should also probably not have `@time` calls in there, but I presume that's temporary for testing purpose.

cpfiffer:
Maybe we need an upstream issue for the language? Can you reduce this problem to show that changing one thing slows things down meaningfully?

test/runtests.jl
devmotion: > they should be nested within an MCMCChains set

Even if that leads to no information about the individual testsets, possibly requires print statements to wake up Travis, and means that tests are not stopped after a failing testset?

> We should also probably not have @time calls in there, but I presume that's temporary for testing purpose.

If everything is nested inside one single testset, the `@time` calls are not useful at all. In the other case, they would make it easier to spot performance regressions and which tests and function calls actually lead to many allocations and bad performance.

devmotion:
No, that's the main problem, so far I haven't been able to figure out which change(s) is causing these performance problems.

test/runtests.jl
cpfiffer: > Even if that leads to no information about the individual testsets, possibly requires print statements to wake up Travis, and means that tests are not stopped after a failing testset?

Yes. Tests should proceed if one test set fails -- we want to know everywhere something fails, not the first place it fails. I'm alright with `println` statements if you want to indicate what's being tested -- Turing's tests have something similar.

I'm fine with leaving in the `@time` statements.

devmotion:
https://github.com/TuringLang/MCMCChains.jl/pull/180 indicates that the problem is not caused by any changes in this PR but rather already present on master. There is quite a significant difference to the [CI tests of master 4 days ago](https://travis-ci.org/TuringLang/MCMCChains.jl/builds/650847167?utm_medium=notification&utm_source=github_status) (e.g., on Julia 1.3 it now takes 26 minutes instead of 10 for them to complete). The only difference seems to be that the new tests use StatsBase 0.32.1, StatsPlots 0.14.0, and TableOperations 0.2.0 instead of StatsBase 0.32.0 and StatsPlots 0.12.0 (TableOperations was just added as a dependency to StatsPlots in the latest release).

devmotion:
Seems some of the latest changes fixed some type inference problems and the CI time outs, so I guess we're getting closer to being able to finish the AbstractMCMC update in Turing :smiley:

PR #180, opened 2020-02-18
Run tests with timings

devmotion:
This PR is not intended to be merged, I just thought it might be good to compare the timings in https://github.com/TuringLang/MCMCChains.jl/pull/179 with the ones on current master.


cpfiffer:
Well that halved testing time for Julia 1.3! I looked through the [diff](https://github.com/JuliaStats/StatsBase.jl/compare/v0.32.0...v0.32.1) for StatsBase and none of the signatures they changed (or even just plain code) suggest something that would cause a performance regression of this magnitude.

devmotion:
Yes, I'm a bit lost here. I'll continue with some debugging and I contacted someone from StatsBase on Slack, but maybe as a workaround we could just upper bound StatsBase to 0.32.0 for now, to be able to move on with the concatenation PR and the AbstractMCMC update in Turing?

Initially, I thought it has to be something about the combination of the new StatsBase release and StatsPlots, due to this massive performance regression in the plotting tests. However, a closer look at the timings reveals that also the "get tests" and the "sampling tests" show performance regressions (timings are increased by a factor of 20 and 4, respectively), whereas all other tests seem to be unaffected.

goedman:
A while back, just after 1.3, I also tried to look at timings on Travis but came to the conclusion Travis resources are probably shared. So sometimes in early weekend the tests would pass, the next morning Or later on they might not. It was not reproducible at all.

devmotion:
Sure, the timings are not completely reproducible, but often they are a good indicator for more significant performance regressions (at least that's my experience with the JuliaDiffEq tests). I'm pretty sure that the regressions that we see here with StatsBase 0.32.1 are real, they are quite severe (from around 12 to 26 minutes) and reproducible.

devmotion:
@cpfiffer What's your opinion? I don't really know how to continue here, and I can't spend much time on debugging the regressions right now. The issue is definitely triggered by the StatsBase update, but I still don't see how. To me it seems, the main underlying issue (throughout this package) are all the type instabilities that lead to absolutely terrible performance in the first run, e.g.,
```julia
julia> using MCMCChains

julia> chn = Chains(rand(1000, 11, 1), [("m[$i]" for i in 1:10)..., "s"]);

julia> @time get_params(chn);
 97.901070 seconds (73.20 M allocations: 4.747 GiB, 2.47% gc time)

julia> @time get_params(chn);
  0.000438 seconds (241 allocations: 100.438 KiB)
```
So maybe the StatsBase update just triggered some additional type inference problems. Should we just continue with the other PR to get AbstractMCMC's update in? Should we bound StatsBase for now until we know what's going on?

cpfiffer:
Let's just get the AbstractMCMC stuff through without bounding StatsBase. The performance issues are separate and need to be addressed in a comprehensive way. Ultimately there's a lot of crap in MCMCChains that needs to be completely refactored. Thanks for poking around this.

PR #181, opened 2020-02-23
Make MCMCChains better

cpfiffer:
Removes a lot of bloat from MCMCChains. This PR should result in a fairly significant speedup for first-time loading.

Changes:
- `DataFrames` is now no longer built in to MCMCChains -- the dataframe creation functionality is now loaded lazily when someone calls `using DataFrames`.
- `ChainDataFrames` are now much slimmer, and use custom text printing tools. They can be accessed largely the same way as the old `ChainDataFrames`. 
- `summarize` is a lot quicker because it doesn't make a `DataFrame` on the backend.

I'll do a rebase shortly to handle the changes in #179.


cpfiffer:
Oh, here's an example of the output:

```julia
Object of type Chains, with data of type 1000×5×3 Array{Float64,3}

Iterations        = 1:1000
Thinning interval = 1
Chains            = 1, 2, 3
Samples per chain = 1000
parameters        = Param1, Param2, Param3, Param4, Param5

2-element Array{ChainDataFrame,1}

Summary Statistics
  parameters    mean     std  naive_se    mcse        ess   r_hat
  ──────────  ──────  ──────  ────────  ──────  ─────────  ──────
      Param1  0.4983  0.2861    0.0052  0.0044  2980.8563  0.9997
      Param2  0.4932  0.2930    0.0053  0.0052  2899.7937  0.9998
      Param3  0.5055  0.2897    0.0053  0.0045  3040.5642  0.9997
      Param4  0.4916  0.2897    0.0053  0.0048  3047.6075  1.0002
      Param5  0.5122  0.2858    0.0052  0.0054  2872.9907  0.9998

Quantiles
  parameters    2.5%   25.0%   50.0%   75.0%   97.5%
  ──────────  ──────  ──────  ──────  ──────  ──────
      Param1  0.0257  0.2571  0.4955  0.7407  0.9739
      Param2  0.0290  0.2370  0.4916  0.7511  0.9768
      Param3  0.0261  0.2582  0.5057  0.7580  0.9782
      Param4  0.0235  0.2323  0.4914  0.7433  0.9722
      Param5  0.0282  0.2676  0.5209  0.7609  0.9726
```

src/constructors.jl
```diff
@@ -66,174 +66,38 @@ function Base.Array(chn::Chains,
...
-                    else
-                        b = hcat(b, reshape(x[Symbol(par)], d*c)[:, 1])
-                    end
-                end
-            end
+    arr = if append_chains
+        if remove_missing_union
+            vcat([convert(Array{Real}, chn.value.data[:,:,i]) for i in 1:size(chn, 3)]...)
+        else
+            vcat([chn.value.data[:,:,i] for i in 1:size(chn, 3)]...)
```
devmotion: Maybe use `mapreduce`?

src/constructors.jl
```diff
@@ -66,174 +66,38 @@ function Base.Array(chn::Chains,
...
-                        else
-                            b[i] = hcat(b[i], x[Symbol(par)][:, i])
-                        end
-                    end
-                end
-            end
+        if remove_missing_union
+            [convert(Array{Real}, chn.value.data[:,:,i]) for i in 1:size(chn, 3)]
+        else
+            [chn.value.data[:,:,i] for i in 1:size(chn, 3)]
```
devmotion: I'm wondering if one should remove the keyword argument `append_chains` and instead use two different functions since the different return types make the function type unstable.

src/constructors.jl
```diff
@@ -66,174 +66,38 @@ function Base.Array(chn::Chains,
...
-                else
-                    if remove_missing_union
-                        b = hcat(b, reshape(convert(Array{Float64}, x[Symbol(par)]), d*c)[:, 1])
-                    else
-                        b = hcat(b, reshape(x[Symbol(par)], d*c)[:, 1])
-                    end
-                end
-            end
+    arr = if append_chains
+        if remove_missing_union
```
devmotion: Are the types not made more concrete by the `map(identity, ...)` calls below, so that `remove_missing_union` can be dropped here?

src/constructors.jl
cpfiffer: What would be a good signature? `jagged_array`? I'm also not a fan of calling this function `Array`, since the actual return value for this should be `chain.value.data`, which is already a 3D array.

Maybe this should be `stacked_array` and `jagged_array`, but I think there's better names we can come up with?

cpfiffer:
@goedman You should take a look at this too, since it would require you to implicity import `DataFrames`.

goedman:
Thanks @cpfiffer , I'm ok with putting it behind the Requires barrier.

Project.toml
```diff
@@ -3,16 +3,18 @@ uuid = "c7f686f2-ff18-58e9-bc7b-31028e88f75d"
 keywords = ["markov chain monte carlo", "probablistic programming"]
 license = "MIT"
 desc = " Chain types and utility functions for MCMC simulations."
-version = "2.0.0"
+version = "3.0.0"
 
 [deps]
 AbstractMCMC = "80f14c24-f653-4e6a-9b94-39d6b0f70001"
 AxisArrays = "39de3d68-74b9-583c-8d2d-e117c070f3a9"
 DataFrames = "a93c6f00-e57d-5684-b7b6-d8193f3e46c0"
```
devmotion: If DataFrames is only optional, it should be moved to the `[extras]` section.

src/constructors.jl
cpfiffer: Let's leave this for now and come back to it later.

cpfiffer:
I'll merge this at the end of the day, unless there's any other comments.

src/MCMCChains.jl
```diff
@@ -4,28 +4,29 @@ using AxisArrays
 const axes = Base.axes
 
 import AbstractMCMC
-using AbstractMCMC: chainscat
-import DataFrames
-using DataFrames: eachcol, DataFrame
+import AbstractMCMC: chainscat
+# import DataFrames
+# import DataFrames: eachcol, DataFrame
```
devmotion: Maybe remove these lines completely?

Project.toml
```diff
@@ -32,6 +34,7 @@ julia = "^1"
 KernelDensity = "5ab0869b-81aa-558d-bb23-cbf5423bbe9b"
 StatsPlots = "f3b207a7-027a-5e70-b257-86293d7955fd"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+DataFrames = "a93c6f00-e57d-5684-b7b6-d8193f3e46c0"
```
devmotion: Adding DataFrames to `[targets]` as well should fix the test errors.

PR #193, opened 2020-03-22
Implement Tables and TableTraits interfaces

sethaxen:
The compatibility code with DataFrames is a bit too specific. We can more generally implement the Tables and TableTraits interfaces, which automatically give compatibility with DataFrames but also any other package that can use that interface, including packages like CSV, VegaLite, StatsPlots, or (soon) StatsMakie.

The interface essentially gives a "tidy" view of the chains (i.e. one draw per row), similar to the DataFrames code. The helper type `TableChains` can be used to reduce the view to specific sections and chains or enforce column types.

If the DataFrames code is removed, this still works:

```julia
julia> DataFrame(chn)
1000×6 DataFrame
│ Row  │ Iteration │ Chain │ Param1    │ Param2   │ Param3  │ Param4  │
│      │ Int64     │ Int64 │ Float64   │ Float64  │ Float64 │ Float64 │
├──────┼───────────┼───────┼───────────┼──────────┼─────────┼─────────┤
│ 1    │ 1         │ 1     │ 1.8934    │ 0.854273 │ 3.84264 │ 1.0     │
│ 2    │ 2         │ 1     │ 2.32506   │ 1.1751   │ 1.16137 │ 2.0     │
│ 3    │ 3         │ 1     │ 1.10374   │ 1.33482  │ 4.40117 │ 1.0     │
│ 4    │ 4         │ 1     │ 2.12994   │ 1.68052  │ 2.06718 │ 2.0     │
│ 5    │ 5         │ 1     │ 2.05949   │ 2.66586  │ 4.20714 │ 2.0     │
│ 6    │ 6         │ 1     │ 1.23058   │ 2.1124   │ 4.07118 │ 2.0     │
⋮
│ 994  │ 494       │ 2     │ 0.0656108 │ 1.18337  │ 2.92848 │ 1.0     │
│ 995  │ 495       │ 2     │ 1.91377   │ 0.515092 │ 2.40351 │ 2.0     │
│ 996  │ 496       │ 2     │ 0.438731  │ 1.30662  │ 3.91179 │ 1.0     │
│ 997  │ 497       │ 2     │ 1.15439   │ 2.24157  │ 1.62304 │ 2.0     │
│ 998  │ 498       │ 2     │ 1.07865   │ 2.35256  │ 3.46465 │ 2.0     │
│ 999  │ 499       │ 2     │ 1.31565   │ 1.33091  │ 4.19777 │ 2.0     │
│ 1000 │ 500       │ 2     │ -1.20609  │ 2.18146  │ 3.84267 │ 2.0     │
```

This is marked as a WIP because it's currently missing tests, code to create a `Chains` from any object that implements the Tables interface, and upper bounds. The main feature it's missing from the current DataFrames code is the ability to globally remove the union with `missing` if possible.


devmotion:
Great! I think we should definitely replace the DataFrames integration with an implementation of the Tables interface.

devmotion:
Having said that, actually I do not believe anymore that a chain should be viewed as a table. At least in many cases that I've seen in Turing and in my own research it is common to have high(er)-dimensional samples (I remember one case where I was dealing with random fields of size 128x128); IMO in these cases reshaping the chain into a table is usually very inefficient and feels a bit weird. I got the feeling that in these cases actually it is more intuitive to view the chain as a set of chains of potentially higher dimensional samples. But I guess the Tables interface is not intended for dealing with columns of both scalar and non-scalar (such as vectors or matrices or ...) entries? Maybe for such cases a representation as StructArray would be helpful?

Nevertheless, I think we should implement the Tables interface.

sethaxen:
> Having said that, actually I do not believe anymore that a chain should be viewed as a table. At least in many cases that I've seen in Turing and in my own research it is common to have high(er)-dimensional samples (I remember one case where I was dealing with random fields of size 128x128); IMO in these cases reshaping the chain into a table is usually very inefficient and feels a bit weird.

I agree in general. And often some of the axes of the variables are shared. Some data structures (e.g. ArviZ's `InferenceData`) take this into account. In general though, most algorithms/visualization packages only handle low- (often one-) dimensional representations of the chains, so having a tabular representation for these packages is important.

mkborregaard:
Doesn't this PR essentially allow for what @devmotion suggests?

devmotion:
> Doesn't this PR essentially allow for what @devmotion suggests?

Don't get me wrong, I'm very much in favor of this PR, but of course it is based on the current design of Chains, and hence every entry in the table has to be a scalar (and even of the same type in the current design!). I was just saying that that's not always a reasonable design or assumption.

> The helper type TableChains can be used to reduce the view to specific sections and chains or enforce column types.

I'm not sure if we actually need that. MCMCChains already supports subsetting the chain, so isn't it sufficient to implement the Tables interface for Chains?

cpfiffer:
> Don't get me wrong, I'm very much in favor of this PR, but of course it is based on the current design of Chains, and hence every entry in the table has to be a scalar (and even of the same type in the current design!). I was just saying that that's not always a reasonable design or assumption.

We can modify this later when we get around to making more structurally typed data forms -- then the Tables interface would still work quite nicely even if you store `NamedTuple`s or something. 

I'm all in favor, thank you for the PR, @sethaxen!

sethaxen:
> > The helper type TableChains can be used to reduce the view to specific sections and chains or enforce column types.
> 
> I'm not sure if we actually need that. MCMCChains already supports subsetting the chain, so isn't it sufficient to implement the Tables interface for Chains?

For the currently supported features, yes, but the DataFrames integration also has a flag to convert arrays of type `Union{Missing,Any}` to `Float64` that defaults to `true`. If that functionality is split out into its own utility function, then there's no use case at the moment for `TableChains`.

> We can modify this later when we get around to making more structurally typed data forms -- then the Tables interface would still work quite nicely even if you store `NamedTuple`s or something.

It's worth noting that some packages will likely always assume that the samples are in a tabular format with real values, especially plotting packages that use a traditional grammar of graphics. Even when we have more structurally typed data forms, having such a table interface will be useful (and maybe this is an argument for keeping the `TableChains` type). I'm thinking such an interface would probably recurse through the types, getting their fields, and then flatten into a table with columns like `Symbol("ParamA.PropertyB.PropertyC")`. And of course we would want an interface as well that does no flattening.

cpfiffer:
Sounds good to me. I'm happy with the status of this. When you get a chance, could you add a couple of tests to hit the main functions in here?

sethaxen:
> Sounds good to me. I'm happy with the status of this. When you get a chance, could you add a couple of tests to hit the main functions in here?

Will do.

Also, with the addition of `concretize` in #194 , there's no need for `TableChains` anymore, so I'll remove it. I'll also remove the DataFrames integration. Since the default behavior of the Tables interface of `Chains` is different from applying the `DataFrame` constructor to `Chains`, this will technically be a breaking change.

sethaxen:
Also, while I'm at it, I'll implement the Tables interface for `ChainDataFrames`.

sethaxen:
Okay, I think this PR is ready to merge.

src/MCMCChains.jl
```diff
@@ -12,11 +12,27 @@ using Formatting
 import StatsBase: autocov, counts, sem, AbstractWeights,
     autocor, describe, quantile, sample, summarystats, cov
 using Requires
+using Tables
+using TableTraits
+using IteratorInterfaceExtensions
```
devmotion: Since you always write Tables.rows etc explicitly (which I like a lot BTW) you should just change it to
```suggestion
import Tables
import TableTraits
import IteratorInterfaceExtensions
```

src/MCMCChains.jl
```diff
@@ -12,11 +12,27 @@ using Formatting
...
+    columns,
+    columnnames,
+    getcolumn,
+    rowaccess,
+    rows,
+    rowtable,
+    namedtupleiterator,
+    schema
+import TableTraits: isiterabletable
+import IteratorInterfaceExtensions: isiterable, getiterator
```
devmotion: For the same reason you don't need these imports:
```suggestion
```

src/tables.jl
```diff
@@ -0,0 +1,87 @@
...
+
+## Chains
+
+Tables.istable(::Type{<:Chains}) = true
+
+Tables.columnaccess(::Type{<:Chains}) = true
+
+Tables.columns(chn::Chains) = chn
+
+Tables.columnnames(chn::Chains) = (:Iteration, :Chain, Symbol.(names(chn))...)
```
devmotion: I'm wondering if we should use `:iteration` and `:chain`?

devmotion: Also, do we have to consider the (not very likely) possibility that one of the names clashes with `:iteration` or `:chain`?

sethaxen: Yeah, I was debating this. The summary dataframe columns use the lower case names, so there's a point in that favor. I ultimately went with upper case since it seemed less likely those would collide with a parameter name in a model. I'm happy to change it to lower case if you like.

src/tables.jl
```diff
@@ -0,0 +1,87 @@
...
+
+function Tables.getcolumn(chn::Chains, i::Int)
+    return Tables.getcolumn(chn, Tables.columnnames(chn)[i])
+end
+function Tables.getcolumn(chn::Chains, nm::Symbol)
+    if nm == :Iteration
+        niter, _, nchains = size(chn)
+        return repeat(Base.OneTo(niter), nchains)
```
devmotion: This should be
```suggestion
        iterations = range(chn)
        nchains = size(chn, 3)
        return repeat(iterations, nchains)
```

src/tables.jl
```diff
@@ -0,0 +1,87 @@
...
+    elseif nm == :Chain
+        chainids = chains(chn)
+        niter = size(chn, 1)
+        return vcat(fill.(chainids, niter)...)
```
devmotion: This should be written as
```suggestion
        return repeat(chainids; inner = niter)
```

src/tables.jl
sethaxen: Oh, I see I was writing my reply as you were writing my comment. There are a few options here: 1) choose a column name that is highly unlikely a user would use (e.g. `:__chain:`) 2) choose a column name based on the variable names (e.g. if a user has `:chain` then use `:chainid`) 3) throw an error if `:chain` is in the parameter names 4) do nothing and hope the user doesn't use one of those names.

I really don't like 1) and 2). 3) is fine. I do think it's unlikely a user would name a parameter `Chain` or `Iteration` if following Julia conventions, but it is certainly possible.

src/tables.jl
```diff
@@ -0,0 +1,87 @@
...
+    else
+        return vec(getindex(chn, nm).value)
```
devmotion: One more reason to get rid of strings internally I guess (as done in https://github.com/TuringLang/MCMCChains.jl/pull/197), since then we could write more efficiently
```julia
        return vec(chn.value[:, nm, :])
```
Unfortunately, at the moment it will error if the parameter names are strings :disappointed:

src/tables.jl
```diff
@@ -0,0 +1,87 @@
...
+Tables.columnaccess(::Type{<:ChainDataFrame}) = true
+
+Tables.columns(cdf::ChainDataFrame) = cdf
+
+Tables.columnnames(cdf::ChainDataFrame) = keys(cdf.nt)
+
+function Tables.getcolumn(cdf::ChainDataFrame, i::Int)
+    return Tables.getcolumn(cdf, keys(cdf.nt)[i])
+end
+Tables.getcolumn(cdf::ChainDataFrame, nm::Symbol) = getproperty(cdf.nt, nm)
```
devmotion: I'm curious, is there a reason for using `getproperty(cdf.nt, nm)` instead of `cdf.nt[nm]`?

src/tables.jl
```diff
@@ -0,0 +1,87 @@
...
+Tables.istable(::Type{<:ChainDataFrame}) = true
+
+Tables.columnaccess(::Type{<:ChainDataFrame}) = true
+
+Tables.columns(cdf::ChainDataFrame) = cdf
+
+Tables.columnnames(cdf::ChainDataFrame) = keys(cdf.nt)
+
+function Tables.getcolumn(cdf::ChainDataFrame, i::Int)
+    return Tables.getcolumn(cdf, keys(cdf.nt)[i])
```
devmotion: Couldn't we just write
```suggestion
Tables.getcolumn(cdf::ChainDataFrame, i::Int) = cdf.nt[i]
```
?

src/tables.jl
sethaxen: Not really, just preference. I usually use the dot syntax with named tuples instead of indexing. I can change if you prefer.

src/tables.jl
```diff
@@ -0,0 +1,87 @@
...
+Tables.rows(cdf::ChainDataFrame) = cdf
+
+Tables.rowtable(cdf::ChainDataFrame) = Tables.rowtable(Tables.columntable(cdf))
+
+function Tables.namedtupleiterator(cdf::ChainDataFrame)
+    return Tables.namedtupleiterator(Tables.columntable(cdf))
+end
+
+function Tables.schema(cdf::ChainDataFrame)
+    return Tables.Schema(keys(cdf.nt), eltype.(values(cdf.nt)))
```
devmotion: I guess we could also write
```suggestion
function Tables.schema(cdf::ChainDataFrame{NamedTuple{names,T}}) where {names,T}
    return Tables.Schema(names, T)
```

src/tables.jl
devmotion: I'm just a bit surprised about the use of `getproperty` with NamedTuples, I would have expected `getfield` rather (which is actually what `getindex` calls in https://github.com/JuliaLang/julia/blob/f449765943ba414bd57c3d1a44a73e5a0bb27534/base/namedtuple.jl#L113). So yes, I think either change it to `getfield` or use standard indexing.

src/tables.jl
sethaxen: Good point, yes, and for that matter `columnnames` could do the same.

src/tables.jl
sethaxen: Ah, I did not know that. Sure I can do that.

src/tables.jl
devmotion: Alternatively one could prefix all parameter names with something like `param_` - but IMO that's pretty bad from a user perspective even though it would solve the issue :stuck_out_tongue:

sethaxen: > Alternatively one could prefix all parameter names with something like `param_` - but IMO that's pretty bad from a user perspective even though it would solve the issue 😛

I agree. I'm not certain what the best approach here is and don't have a preference. I'm warming up to 1), but I'll defer to others with more of a preference.

devmotion: Actually, I was leaning towards 3), since if a user used `:chain` or `:iteration` as parameter name, she could just change it with `set_names` or in the upstream code (and I strongly believe there shouldn't be any "magic" renaming in the case of name clashes). But how could we reliably throw an error in 3)? A user (or developer) doesn't have (?) to call `columnnames`, but `getcolumn` would still be problematic. So if we want to throw an error, at which stage should this be done?

test/tables_tests.jl
```diff
@@ -0,0 +1,134 @@
...
+            nt = Tables.rowtable(chn)[2]
+            @test nt ==
+                  (; (k => Tables.getcolumn(chn, k)[2] for k in Tables.columnnames(chn))...)
+            @test nt == collect(Iterators.take(Tables.namedtupleiterator(chn), 2))[2]
+            @test Tables.schema(chn) isa Tables.Schema
+            @test Tables.schema(chn).names ===
+                  (:Iteration, :Chain, :a, :b, :c, :d, :e, :f, :g, :h)
+            @test Tables.schema(chn).types === (
+                Int64,
+                Int64,
```
devmotion: Better to change it to
```suggestion
                Int,
                Int,
```
otherwise the tests fail on 32bit.

test/tables_tests.jl
```diff
@@ -0,0 +1,134 @@
...
+        chn = Chains(val, colnames, Dict(:internals => internal_colnames))
+        cdf = describe(chn)[1]
+
+        @testset "Tables interface" begin
+            @test Tables.istable(typeof(cdf))
+            @test Tables.columnaccess(typeof(cdf))
+            @test Tables.columns(cdf) === cdf
+            @test Tables.columnnames(cdf) == keys(cdf.nt)
+            @testset for k in keys(cdf.nt)
+                @test Tables.getcolumn(cdf, k) == getproperty(cdf.nt, k)
```
devmotion: I'd suggest avoiding `getproperty` in the tests as well:
```suggestion
            for (k, v) in pairs(cdf.nt)
                @test Tables.getcolumn(cdf, k) == v
```

test/tables_tests.jl
```diff
@@ -0,0 +1,134 @@
...
+            end
+            @test Tables.getcolumn(cdf, 1) == Tables.getcolumn(cdf, keys(cdf.nt)[1])
+            @test Tables.getcolumn(cdf, 2) == Tables.getcolumn(cdf, keys(cdf.nt)[2])
+            @test_throws Exception Tables.getcolumn(cdf, :blah)
+            @test_throws Exception Tables.getcolumn(cdf, length(keys(cdf.nt)) + 1)
```
devmotion:
```suggestion
            @test_throws Exception Tables.getcolumn(cdf, length(cdf.nt) + 1)
```

test/tables_tests.jl
```diff
@@ -0,0 +1,134 @@
...
+            @test Tables.rowaccess(typeof(cdf))
+            @test Tables.rows(cdf) === cdf
+            @test length(Tables.rowtable(cdf)) == length(values(cdf.nt)[1])
```
devmotion:
```suggestion
            @test length(Tables.rowtable(cdf)) == length(cdf.nt[1])
```

test/tables_tests.jl
```diff
@@ -0,0 +1,134 @@
...
+            @test Tables.columntable(cdf) == cdf.nt
+            nt = Tables.rowtable(cdf)[1]
+            @test nt == (; (k => getproperty(cdf.nt, k)[1] for k in keys(cdf.nt))...)
```
devmotion:
```suggestion
            @test nt == (; (k => v[1] for (k, v) in pairs(cdf.nt))...)
```

test/tables_tests.jl
```diff
@@ -0,0 +1,134 @@
...
+            @test nt == collect(Iterators.take(Tables.namedtupleiterator(cdf), 1))[1]
+            nt = Tables.rowtable(cdf)[2]
+            @test nt == (; (k => getproperty(cdf.nt, k)[2] for k in keys(cdf.nt))...)
```
devmotion:
```suggestion
            @test nt == (; (k => v[2] for (k, v) in pairs(cdf.nt))...)
```

test/tables_tests.jl
```diff
@@ -0,0 +1,134 @@
...
+            @test Tables.schema(cdf) isa Tables.Schema
+            @test Tables.schema(cdf).names === keys(cdf.nt)
+            @test Tables.schema(cdf).types === eltype.(values(cdf.nt))
+        end
+
+        @testset "TableTraits interface" begin
+            @test IteratorInterfaceExtensions.isiterable(cdf)
+            @test TableTraits.isiterabletable(cdf)
+            nt = collect(Iterators.take(IteratorInterfaceExtensions.getiterator(cdf), 1))[1]
+            @test nt == (; (k => getproperty(cdf.nt, k)[1] for k in keys(cdf.nt))...)
```
devmotion:
```suggestion
            @test nt == (; (k => v[1] for (k, v) in pairs(cdf.nt))...)
```

test/tables_tests.jl
```diff
@@ -0,0 +1,134 @@
...
+            nt = collect(Iterators.take(IteratorInterfaceExtensions.getiterator(cdf), 2))[2]
+            @test nt == (; (k => getproperty(cdf.nt, k)[2] for k in keys(cdf.nt))...)
```
devmotion:
```suggestion
            @test nt == (; (k => v[2] for (k, v) in pairs(cdf.nt))...)
```

src/tables.jl
sethaxen: I think the right way is to perform the check in `Tables.rows` and `Tables.columns`, as well as `Tables.schema` to be safe, since these are the [core interface functions](https://juliadata.github.io/Tables.jl/stable/#Implementing-the-Interface-(i.e.-becoming-a-Tables.jl-source)-1) and are recommended to be called before calling functions like `getcolumn`. I'll go ahead and implement that as well as name changes to `:chain` and `:iteration` tomorrow.

devmotion: Sounds good, thanks!

sethaxen: Ended up implementing it tonight.

src/tables.jl
```diff
@@ -2,23 +2,33 @@
 
 ## Chains
 
+function _check_columnnames(chn::Chains)
+    colnames = Symbol.(names(chn))
+    for nm in (:iteration, :chain)
+        nm in colnames && error("'$(nm)' is a reserved column name. Rename the parameter.")
+    end
```
devmotion: `Symbol.(names(chn))` always creates a new array of parameter names, even if the parameter names are already symbols. I'd suggest:
```suggestion
    for name in names(chn)
        symname = Symbol(name)
        if symnname === :iteration || symnname === :chains
            error("'$(name)' is a reserved column name. Please rename the parameter.")
        end
    end
```

src/tables.jl
```diff
@@ -0,0 +1,102 @@
...
+Tables.rows(cdf::ChainDataFrame) = cdf
+
+Tables.rowtable(cdf::ChainDataFrame) = Tables.rowtable(Tables.columntable(cdf))
+
+function Tables.namedtupleiterator(cdf::ChainDataFrame)
+    return Tables.namedtupleiterator(Tables.columntable(cdf))
+end
+
+function Tables.schema(::ChainDataFrame{NamedTuple{names,T}}) where {names,T}
+    return Tables.Schema(names, eltype.(T.parameters))
```
devmotion: A bit easier (to read and for the compiler) might be
```suggestion
    return Tables.Schema(names, eltype.(fieldtypes(T)))
```

devmotion:
Thanks a lot, it looks good! The Travis CI test errors on Julia 1.0 and 1.1 seem to be unrelated (for whatever reason Pkg tries to install a package from https://github.com/SciML/DiffEqDiffTools.jl which does not exist or is private). I think we can merge as soon as we figure out what's going on there and tests pass.

sethaxen:
Yeah, I'm currently trying to chase down what the culprit package is. The JuliaDiffEq organization just changed to SciML, but that doesn't fully explain this.

On Slack RE DiffEqDiffTools, @ChrisRackauckas said:
> ...that repo became  JuliaDiff/FiniteDiff.jl a few months ago: https://discourse.julialang.org/t/finitediff-jl-fast-sparse-gradients-jacobians-hessians/32976

> what package is still requiring DiffEqDiffTools? The right thing to do would be to update that.

devmotion:
OK, so apparently other people have the same CI issues. DiffEqDiffTools became FiniteDiff quite a while ago, but apparently the renaming of JuliaDiffEq to SciML causes problems now.

devmotion:
Ah yeah I just read that :smile:

ChrisRackauckas:
What package requires DiffEqDiffTools? I'll go put the PR in myself to update it: there's literally no change to the API.

sethaxen:
I don't know yet. Looks like even in the last build that succeeded (before the org change I assume), DiffEqDiffTools was being pulled in: https://travis-ci.com/github/TuringLang/MCMCChains.jl/jobs/302769419#L472

sethaxen:
Ah okay, looks like it's because an old version of NLSolversBase is being pulled in (v7.5.0, which requires DiffEqDiffTools). Now to figure out why that is being pulled in.

sethaxen:
Ah, there it is, it's because NLSolversBase > v7.6.0 requires Julia 1.2 or higher. Hence why Julia 1.0 and 1.1 fail for us. Not certain what the solution here is.

ChrisRackauckas:
@pkofod NLSolversBase patch release which changes to FiniteDiff?

devmotion:
Is it possible to fix the registry? It's pointing to the nonexisting repo, so even people that want to install DiffEqDiffTools can't do it anymore. Would it be sufficient just to change the link in Package.toml to FiniteDiff? I guess before Github was redirecting the link automatically but after the rename that's not the case anymore.

ChrisRackauckas:
Good idea. https://github.com/JuliaRegistries/General/pull/11578 would do it.

ChrisRackauckas:
Rerun CI and it should be fixed.

sethaxen:
Works! Thanks!

ChrisRackauckas:
Awesome.

Issue #189, opened 2020-03-02
Cor no longer works


itsdfish:
I found that cor no longer works. Here is a MWE:

```
using MCMCChains, StatsPlots 
import MCMCChains: cor 

val = rand(500,5, 2)
chn = Chains(val, ["a", "b", "c", "d", "e"])
cor(chn)
```
Error:
```
UndefVarError: DataFrame not defined
#cor#193(::Bool, ::Bool, ::Array{Symbol,1}, ::Int64, ::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}, ::typeof(cor), ::Chains{Float64,Missing,NamedTuple{(:parameters,),Tuple{Array{Any,1}}},NamedTuple{(),Tuple{}}}) at stats.jl:59
cor(::Chains{Float64,Missing,NamedTuple{(:parameters,),Tuple{Array{Any,1}}},NamedTuple{(),Tuple{}}}) at stats.jl:59
top-level scope at untitled-84e45c9bd5bfc567b46b3d7282fc1695:5
```

```
(v1.3) pkg> st MCMCChains
    Status `~/.julia/environments/v1.3/Project.toml`
  [31c24e10] Distributions v0.22.5
  [c7f686f2] MCMCChains v3.0.3
  [2913bbd2] StatsBase v0.32.1
```

devmotion:
The problem is that `DataFrame` is still used in `src/stats.jl` even though DataFrames is only an optional dependency. You should be able to work around the issue by explicitly loading DataFrames, but of course this should be fixed properly.

itsdfish:
Thanks. Unfortunately, loading DataFrames failed as a workaround. It produced the same error message.

devmotion:
Ah, I forgot, one should run `@eval MCMCChains begin using DataFrames end`. However, unfortunately that throws a different error since the constructor of `ChainDataFrame` was changed.

devmotion:
This was fixed by https://github.com/TuringLang/MCMCChains.jl/pull/194.

PR #194, opened 2020-03-23
Add `concretize` and fix https://github.com/TuringLang/MCMCChains.jl/issues/189

devmotion:
This PR contains the following changes:
- It adds an unexported `concretize` method that can be used to transform a `Chains`, `AbstractDataFrame`, or (possibly nested) `AbstractArray` with element types that are not concretely typed (such as `Real` or `Union{Missing,Float64}`) to a version with concretely typed element types if possible (clearly it can be type stable only if all element types are already concretely type in which case it just returns the input)
- It adds separate functions to transform a `Chains` to a matrix, a vector of scalars, a vector of vectors, and a vector of matrices and makes use of these functions in the `Array` constructor
- It removes an abstract field in `Chains` by adding an additional type parameter, which, e.g., makes the functions in the last point inferrable and type stable (at least for the arguments that are used in the tests)

A side remark:
I'm not happy about the `Array` and `DataFrame` constructors since it it is impossible to infer their return types and they can't be type stable. Actually, I think we should consider removing them and instead export functions that return ALWAYS vectors of matrices (i.e., matrices for each chain) or matrices (i.e., with all chains appended). In particular, I think we should stop dropping the last dimension if there is only one parameter (this conversion/reshaping should only happen downstream if it is needed, similar to the behaviour of `sum` etc.), and we should remove most of the keyword arguments - IMO instead of always providing `section` keyword arguments users should just compute the subset of the chain they are interested in before hand, they should sort the chain by using the sorting functionality, and they should concretize the chain outside of these functions (ideally they know which types to expect, so they can perform some proper conversion instead of using the type unstable `concretize` approach).

### DIFF
--- a/src/heideldiag.jl
+++ b/src/heideldiag.jl
@@ -33,6 +33,7 @@ function heideldiag(chn::Chains;
                     sections::Vector{Symbol}=[:parameters],
                     showall=false,
                     sorted=true,
+                    digits=4,
                     args...
                    )
     c = showall ?
@@ -56,16 +57,23 @@ function heideldiag(chn::Chains;
                            )
     end
 
-    colnames = tuple(Symbol.(["parameters", "Burn-in", "Stationarity", "p-value", "Mean",
-        "Halfwidth", "Test"])...)
+    # Retrieve columns.
+    columns = [[vals[k][:, i] for i in 1:6] for k in 1:m]
 
-    # Round values.
-    pnames = Symbol.(names(c))
-    vals = map(x -> round.(x, digits=4), vals)
-    columns = [vcat([pnames], [vals[k][:,i] for i in 1:6]) for k in 1:m]
+    # Obtain names of parameters.
+    names_of_params = names(chn)
 
-    dfs = [NamedTuple{colnames}(tuple(columns[k]...)) for k in 1:m]
-    dfs_wrapped = [ChainDataFrame("Heidelberger and Welch Diagnostic - Chain $k",
-                   dfs[k]) for k in 1:m]
-    return dfs_wrapped
+    # Compute data frames.
+    vector_of_df = [
+        ChainDataFrame(
+            "Heidelberger and Welch Diagnostic - Chain $i",
+            (parameters = names_of_params, var"Burn-in" = column[1],
+             Stationarity = column[2], var"p-value" = column[3], Mean = column[4],
+             Halfwidth = column[5], Test = column[6]);
+             digits = digits
+        )
+        for (i, column) in enumerate(columns)
+    ]
+
+    return vector_of_df
 end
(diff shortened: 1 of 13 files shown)


devmotion:
I added some fixes for https://github.com/TuringLang/MCMCChains.jl/issues/189 and fixed `changerate` as well.

devmotion:
Maybe I should revert the change of the type signature

devmotion:
Then we could make this into a bugfix release, and change the signature in another PR.

src/heideldiag.jl
```diff
@@ -56,16 +57,23 @@ function heideldiag(chn::Chains;
...
-    dfs = [NamedTuple{colnames}(tuple(columns[k]...)) for k in 1:m]
-    dfs_wrapped = [ChainDataFrame("Heidelberger and Welch Diagnostic - Chain $k",
-                   dfs[k]) for k in 1:m]
-    return dfs_wrapped
+    # Compute data frames.
+    vector_of_df = [
+        ChainDataFrame(
+            "Heidelberger and Welch Diagnostic - Chain $i",
+            (parameters = names_of_params, var"Burn-in" = column[1],
+             Stationarity = column[2], var"p-value" = column[3], Mean = column[4],
```
cpfiffer:
```suggestion
            (parameters = names_of_params, burn_in = column[1],
             stationarity = column[2], p_value = column[3], mean = column[4],
```

`var". . . "` isn't available prior to Julia 1.3. I think we should probably just use lowercase names anyway.

src/rafterydiag.jl
```diff
@@ -77,15 +78,22 @@ function rafterydiag(
...
-    dfs_wrapped = [ChainDataFrame("Raftery and Lewis Diagnostic - Chain $k",
-                   dfs[k]) for k in 1:m]
-    return dfs_wrapped
+    # Compute data frames.
+    vector_of_df = [
+        ChainDataFrame(
+            "Raftery and Lewis Diagnostic - Chain $i",
+            (parameters = names_of_params, Thinning = column[1], var"Burn-in" = column[2],
+             Total = column[3], Nmin = column[4], var"Dependence Factor" = column[5]);
+            digits = digits
```
cpfiffer:
```suggestion
            (parameters = names_of_params, Thinning = column[1], burn_in = column[2],
             total = column[3], n_min = column[4], dependence_factor = column[5]);
            digits = digits
```

Lowercase, no `@var_str`. See above.

src/heideldiag.jl
devmotion: Yes, I just noticed that it won't work with older Julia versions. Should we maybe use `pvalue` and `burnin` instead of `burn_in` and `p_value`?

cpfiffer:
If that seems reasonable, sure. The changes I recommended above would also go into the signature change above, since they are potentially breaking.

devmotion:
I'll include these changes in a separate PR.

cpfiffer:
Thanks!

PR #199, opened 2020-03-26
Add compat for Tables/TableTraits

sethaxen:
Forgot to include version bounds in #193 .

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -27,10 +27,13 @@ AbstractMCMC = "0.4, 0.5"
 AxisArrays = "^0.3, 0.4"
 Distributions = "^0.21, 0.22, 0.23"
 Formatting = "0.4"
+IteratorInterfaceExtensions = "0.1.1, 1"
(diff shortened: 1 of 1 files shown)


devmotion:
Are you sure that it works with the old versions and do you think compatibility is needed? I'm wondering since Travis just runs test with the most recent versions.

sethaxen:
> Are you sure that it works with the old versions and do you think compatibility is needed? I'm wondering since Travis just runs test with the most recent versions.

Pretty sure. I tested locally with the old versions of TableTraits/IteratorInterfaceExtensions, and tests passed. Also, the documentation RE how to implement the interface did not change from the old to new versions. Finally, to identify the functions to overload, I used as my reference DataFrames, which also indicates compatibility with the same older versions.

devmotion:
OK, perfect!

devmotion:
Thanks a lot!

sethaxen:
You're welcome!

PR #206, opened 2020-04-02
Fix upper and lower bounds of HPD

itsdfish:
The upper and lower bounds were reversed in the hpd output. This PR fixes that issue.


devmotion:
LGTM but at some point we should rewrite the implementation, it's a bit stupid to recompute everything twice.

PR #211, opened 2020-04-14
Remove `showall` and reintroduced `digits` keyword arguments

devmotion:
This PR removes the `showall` keyword argument, instead one has to specify `sections = nothing` now. Moreover, somehow the previously merged PR reintroduced some `digits` keyword arguments which are removed in this PR as well.

### DIFF
--- a/src/constructors.jl
+++ b/src/constructors.jl
@@ -19,7 +19,7 @@ function sort_sections(chn::Chains)
 end
 
 """
-    Array(chains[, sections = :parameters;
+    Array(chains[, sections;
           append_chains = true, remove_missing_union = true])
 
 Construct an `Array` from a chain.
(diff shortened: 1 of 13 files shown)


src/constructors.jl
```diff
@@ -19,58 +19,32 @@ function sort_sections(chn::Chains)
...
 * `Array(chns, append_chains=false)`  : Array of Arrays, each chain in its own array
 * `Array(chns, remove_missing_union=false)` : No conversion to remove missing values
-```
-
 """
-function Base.Array(chain::Chains,
-    sections::Union{Symbol, Vector{Symbol}}=Symbol[:parameters];
+function Base.Array(
+    chains::Chains,
+    sections = :parameters;
```
cpfiffer, line 42: It occurs to me that this should probably be `:parameters` if `:parameters in keys(chain.name_map)`, because we can have cases where subsetting gets us a chain without a `:parameter` key.

devmotion: But this problem can occur for any method that allows to specify a `sections` argument that defaults to `:parameters`, can't it? But since we use `_clean_sections` that would not error but just lead to an empty chain in these cases, it seems?

cpfiffer: True. Should we switch to `nothing` in this case to return the whole chain? Or is it sufficient to return an empty chain? I think returning an empty chain might be confusing if you had done something like `describe(chain[:lp])` or something, where it's clear you only want to see `:lp` but it's locked behind `:internal`. Or maybe this is already happening on the backend.

devmotion: `describe(chain[:lp])` won't work since `chain[:lp]` returns an AxisArray (see https://github.com/TuringLang/MCMCChains.jl/pull/197), but `describe(chain[[:lp]])` would be problematic. I'm not sure what the best (less confusing and most consistent) solution is in this case, maybe one could define
```julia
_default_sections(chains::Chains) = haskey(chains.name_map, :parameters) ? :parameters : nothing
```
and use that as default value for `sections`. Since `name_map` is a NamedTuple that can be inferred as well.

cpfiffer: That's probably sufficient.

PR #156, opened 2019-12-04
[WIP] Implement multiple ESS algorithms

devmotion:
Sorry for this rather incomplete and unfinished PR, but I realized that I probably won't have time to finish it before leaving to NeurIPS and I guess there are many things to discuss, so I just post a draft, a benchmark, and some comments.

This PR provides 3 different implementations of the effective sample size. The first two (`ESSMethod` and `FFTESSMethod`) are very similar to the implementation in Stan, basically the (only?) difference is that this PR doesn't use Stan's improved estimate for the antithetic case (https://github.com/stan-dev/stan/blob/0281899d26644444384ba11f55fda161d52daa18/src/stan/analyze/mcmc/compute_effective_sample_size.hpp#L119-L122). `ESSMethod` estimates the autocorrelation for each desired lag separately in a sequential manner, whereas `FFTESSMethod` uses a FFT to compute the autocorrelation for all lags "simultaneously" (as Stan does as well). In contrast to these two methods, which are based on the plug-in estimator of the autocorrelation and the discussion in https://arxiv.org/pdf/1903.08008.pdf, the third method `BDAESSMethod` uses the (currently implemented) variogram-based estimation of the autocorrelation, as described, e.g., in the Bayesian Data Analysis book. All methods use the identical way of estimating split-Rhat.

I'm still a bit unsatisfied with the current draft since I think preferably algorithms such as ESS estimation should be implemented in a way that is quite agnostic about the structure of the chains and is not particularly tailored towards `Chains` (as a side remark, I think that holds for most diagnostic tools; it seems a bit unfortunate that, even though implemented for `AbstractChains`, most methods seem to assume a `Chains`-like structure - however, maybe that's an intentional design choice). If the design goal is to have only specific algorithms for `Chains` in this package, maybe generic algorithms should be added to a separate package?

Moreover, in contrast to the current implementation `missing` values lead to `missing` values of the ESS and Rhat. If this should be changed, the implementation has to be modified quite a bit since currently it is assumed and exploited that all chains are of the same length (as in Stan's implementation and most papers I've seen). I doubt that the current implementation handles chains of different sizes (which might result from skipping `missing` values) correctly: the number of samples `N` (in Stan's notation) is just set to the minimum number of samples in a chain for the first parameter (although it could differ for different parameters), and used in the further calculations even though the between-chain and within-chain variances are estimated from all available samples. Intuitively, the calculations should be adjusted and account for the fact that there might be different samples per chain?

[...]

PR #215, opened 2020-05-04
ESS algorithms

denainjs:
This PR builds on [PR #156](https://github.com/TuringLang/MCMCChains.jl/pull/156). I modified `devmotion:ess` in order to make it compatible with `TuringLang:dev`.

As in the previous PR, if the chain contains `missing` values for a parameter, then the `rhat` and `ess` for this parameters are automatically set to `missing`.

Benchmarking: for the same example as [PR #156](https://github.com/TuringLang/MCMCChains.jl/pull/156), I obtain:

- `25.437 ms (8388 allocations: 31.78 MiB)` by default
- `25.535 ms (8389 allocations: 31.78 MiB)` explicitly specifying the default method `ESSMethod`
- `288.771 ms (1124 allocations: 34.50 MiB)` for `FFTMethod`
- `89.667 ms (1029 allocations: 31.33 MiB)` for `BDAMethod`

This is my first PR with Turing, let me know if there's anything I'm missing :)


Project.toml
```diff
@@ -6,9 +6,11 @@ desc = "Chain types and utility functions for MCMC simulations."
 version = "4.0.0-DEV"
 
 [deps]
+AbstractFFTs = "621f4979-c628-5d54-868e-fcf4e3e8185c"
 AbstractMCMC = "80f14c24-f653-4e6a-9b94-39d6b0f70001"
 AxisArrays = "39de3d68-74b9-583c-8d2d-e117c070f3a9"
 Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
+FFTW = "7a1cc6ca-52ef-59f5-83cd-3a7055c09341"
```
devmotion, line 13: I think we don't need to (and actually don't want to) include `FFTW` as dependency. Everything can be implemented using the interface in `AbstractFFTs`, and then users can load `FFTW` (or an alternative implementing the interface) if they want to use the FFT method.

devmotion:
```suggestion
```

Project.toml
```diff
@@ -24,9 +26,11 @@ TableTraits = "3783bdb8-4a98-5b6b-af9a-565f29a5fe9c"
 Tables = "bd369af6-aec1-5ad0-b16a-f7cc5008161c"
 
 [compat]
+AbstractFFTs = "0.5"
 AbstractMCMC = "0.4, 0.5"
 AxisArrays = "0.3, 0.4"
 Distributions = "0.21, 0.22, 0.23"
+FFTW = "1.2.1"
```
devmotion, line 33: If it's not a dependency, then we don't need a compatibility bound.
```suggestion
```

Project.toml
devmotion: BTW it's fine (and necessary) to include `FFTW` as test dependency.

src/MCMCChains.jl
```diff
@@ -2,10 +2,11 @@ module MCMCChains
 
 using AxisArrays
 const axes = Base.axes
-
+using AbstractFFTs
 import AbstractMCMC
-import AbstractMCMC: chainscat
+import AbstractMCMC: chainscat, AbstractChains
 using Distributions
+using FFTW
```
devmotion, line 9: See above:
```suggestion
```

src/ess.jl
```diff
@@ -0,0 +1,353 @@
...
+struct BDAESSMethod <: AbstractESSMethod end
+
+# caches
+mutable struct ESSCache{T}
+    samples::Matrix{T}
+    var::Vector{T}
+end
+
+mutable struct FFTESSCache{T,P,I}
+	A::T
```
devmotion, line 52: Was my mistake (shouldn't have used tab):
```suggestion
    A::T
```

src/ess.jl
```diff
@@ -0,0 +1,353 @@
...
+end
+
+mutable struct BDAESSCache{T}
+    samples::Matrix{T}
+    var::Vector{T}
+end
+
+function build_cache(::ESSMethod, samples::Matrix, var::Vector)
+	# check arguments
+	niter, nchains = size(samples)
```
devmotion, line 65:
```suggestion
    # check arguments
    niter, nchains = size(samples)
```

src/ess.jl
```diff
@@ -0,0 +1,353 @@
...
+end
+
+function build_cache(::FFTESSMethod, samples::Matrix, var::Vector)
+	# check arguments
+	niter, nchains = size(samples)
+	length(var) == nchains || throw(DimensionMismatch())
+
+	# create cache for FFT
+	T = complex(eltype(samples))
+	n = nextprod([2, 3], 2 * niter - 1)
```
devmotion, line 78:
```suggestion
    # check arguments
    niter, nchains = size(samples)
    length(var) == nchains || throw(DimensionMismatch())

    # create cache for FFT
    T = complex(eltype(samples))
    n = nextprod([2, 3], 2 * niter - 1)
```

src/ess.jl
```diff
@@ -0,0 +1,353 @@
...
+    A = Matrix{T}(undef, n, nchains)
+
+	# create plans of FFTs
+	fft_plan = plan_fft!(A, 1)
+	ifft_plan = plan_ifft!(A, 1)
+
+	return FFTESSCache(A, fft_plan, ifft_plan, niter)
```
devmotion, line 85:
```suggestion
    # create plans of FFTs
    fft_plan = plan_fft!(A, 1)
    ifft_plan = plan_ifft!(A, 1)

    return FFTESSCache(A, fft_plan, ifft_plan, niter)
```

src/ess.jl
```diff
@@ -0,0 +1,353 @@
...
+end
+
+function build_cache(::BDAESSMethod, samples::Matrix, var::Vector)
+	# check arguments
+	nchains = size(samples, 2)
```
devmotion, line 90:
```suggestion
    # check arguments
    nchains = size(samples, 2)
```

src/ess.jl
```diff
@@ -0,0 +1,353 @@
...
+    length(var) == nchains || throw(DimensionMismatch())
+
+    return BDAESSCache(samples, var)
+end
+
+update_cache!(cache, samples::Matrix, var::Vector) = nothing
+
+function update_cache!(cache::FFTESSCache, samples::Matrix, var::Vector)
+	# check arguments
```
devmotion, line 99:
```suggestion
    # check arguments
```

src/ess.jl
```diff
@@ -0,0 +1,353 @@
...
+			A[i, j] = zero(T)
+		end
+	end
+
+	# compute unnormalized autocorrelation
+	cache.plan * A
+	@. A = abs2(A)
+	cache.invplan * A
+
+	nothing
```
devmotion, line 122:
```suggestion
    @inbounds for j in 1:nchains
        for i in 1:niter
            A[i, j] = samples[i, j]
        end
        for i in (niter + 1):n
            A[i, j] = zero(T)
        end
    end

    # compute unnormalized autocorrelation
    cache.plan * A
    @. A = abs2(A)
    cache.invplan * A

    nothing
```

src/ess.jl
```diff
@@ -0,0 +1,353 @@
...
+
+# estimation of the autocorrelation function
+function mean_autocorr(k::Int, cache::ESSCache)
+    # check arguments
+    samples = cache.samples
+    niter, nchains = size(samples)
+    0 ≤ k < niter || throw(ArgumentError("only lags ≥ 0 and < $niter are supported"))
+
+    # compute mean autocorrelation
+	var = cache.var
```
devmotion, line 133:
```suggestion
    var = cache.var
```

src/ess.jl
```diff
@@ -0,0 +1,353 @@
...
+end
+
+function mean_autocorr(k::Int, cache::BDAESSCache)
+    # check arguments
+    samples = cache.samples
+    niter, nchains = size(samples)
+    0 ≤ k < niter || throw(ArgumentError("only lags ≥ 0 and < $niter are supported"))
+
+    # compute mean autocorrelation
+	var = cache.var
```
devmotion, line 176:
```suggestion
    var = cache.var
```

src/ess.jl
```diff
@@ -0,0 +1,353 @@
...
+            sj += abs2(samples[i, j] - samples[k + i, j])
+        end
+        s += sj / var[j]
+    end
+
+    return 1 - s / (2 * length(samples))
+end
+
+"""
+    ess(chains::Chains; sections = _default_sections(chains), kwargs...)
```
devmotion, line 190: I'm not sure if we have to mention the `sections` keyword argument explicitly in the docstring.

src/ess.jl
```diff
@@ -0,0 +1,353 @@
...
+
+Estimate the effective sample size and the potential scale reduction.
+"""
+function ess(
+    chains::Chains;
+    sections = _default_sections(chains), # perhaps directly nothing by default - or no sections argument?
```
devmotion, line 196: `_default_sections` is mainly for cases where there is no `:parameters` sections in the chain. In the current design of `Chains` (which might change or maybe be extended by an alternate implementation), both the samples and additional information such as log probability and step size are saved in one array but can be distinguished by putting them into different sections. So the logic behind `_default_sections` is that as a default we only want to calculate the ESS for the `:parameters` section if it exists but if there exists no `:parameters` section we want to calculate it for all sections.

src/ess.jl
```diff
@@ -0,0 +1,353 @@
...
+    kwargs...
+)
+    # subset the chain
+    _chains = Chains(chains, _clean_sections(chains, sections)) # may be unnecessary, already applied upstream
```
devmotion, line 200: Users are supposed to be able to call the method directly, hence if we want to provide the functionality to compute the ESS only for specific sections we have to compute the subset here. That being said, if users want they can still write efficient code by subsetting the chain manually and providing `sections = nothing`. In this case `Chains(chains, _clean_sections(chains, sections))` just returns `chains`. BTW in internal methods we should do the same if possible (probably not done currently) and subset the chain at the earliest possible point and then just pass around `sections = nothing`.

src/ess.jl
```diff
@@ -0,0 +1,353 @@
...
+
+    # define cache for the computation of the autocorrelation
+    esscache = build_cache(method, samples, chain_var)
+ 
+    # define output arrays
+    ess = Vector{T}(undef, nparams)
+    rhat = Vector{T}(undef, nparams)
+
+    # for each parameter
+    for (i, chains_slice) in enumerate(eachslice(chains; dims = 2))
```
devmotion, line 246: My fault but I think
```suggestion
    for (i, chains_slice) in enumerate(eachcol(chains))
```
is easier to read.

src/ess.jl
```diff
@@ -0,0 +1,353 @@
...
+    nparams = size(chains, 2)
+    nchains = 2 * size(chains, 3)
+    ntotal = niter * nchains
+
+    # do not compute estimates if there is only one sample or lag
+    maxlag = min(maxlag, niter - 1)
+    maxlag > 0 || return fill(missing, nparams), fill(missing, nparams)
+
+    # define caches for mean and variance
+    U = typeof(float(zero(eltype(chains))))
```
devmotion, line 229: I suggest
```suggestion
    U = typeof(zero(eltype(chains)) / 1)
```
instead of converting to floating point numbers. It's the logic in Statistics and, e.g., it avoids getting `Float64` when operating with `Rational`.

src/ess.jl
```diff
@@ -0,0 +1,353 @@
...
+    T = eltype(chains) >: Missing ? Union{Missing,U} : U
```
devmotion, line 230: Maybe I have written this, but I suggest the more general
```suggestion
    T = promote_type(eltype(chains), typeof(zero(eltype(chains)) / 1))
```

src/ess.jl
devmotion: See my comment below for an approach that also handles `missing`.

src/stats.jl
```diff
@@ -379,8 +258,8 @@ function summarystats(
     # Subset the chain.
     _chains = Chains(chains, _clean_sections(chains, sections))
 
-    # Caluclate ESS separately.
-    ess_df = ess(_chains; sections = nothing, maxlag = maxlag)
+    # Calculate ESS separately.
+    ess_df = ess(_chains; sections = nothing, maxlag = maxlag) # this is a ChainDataFrame - maybe sections = _default_sections() by default
```
devmotion, line 262: Ah we actually already pass `sections = nothing` :smiley:

devmotion: Regarding your comment, why would you want to use `sections = default_sections()` instead?

src/stats.jl
```diff
@@ -219,145 +219,24 @@ function quantile(
...
+    )
 
 Compute the mean, standard deviation, naive standard error, Monte Carlo standard error,
 and effective sample size for each parameter in the chain.
 
 Setting `append_chains=false` will return a vector of dataframes containing the summary
 statistics for each chain.
 
-The remaining keyword arguments are passed to the `msce` function.
+maxlag is the maximum lag for which autocorrelations can be computed
```
devmotion, line 239:
```suggestion
When estimating the effective sample size, autocorrelations are computed for at most `maxlag` lags.
```

test/ess_tests.jl
```diff
@@ -0,0 +1,88 @@
+using MCMCChains
```
devmotion, line 1: Maybe that would be a good time to structure the tests in a similar way as in other Turing packages by having one test file of the same name for each source file (i.e., `test/ess.jl` in this case). We should structure the source files better (in a separate PR), then we could do the same also for other tests.

test/ess_tests.jl
```diff
@@ -0,0 +1,88 @@
...
+        # analyze chain
+        ess_df = ess(chain; method = method)
+
+        # analyze array
+        ess_array, rhat_array = MCMCChains.ess_rhat(x; method = method)
+
+        @test ess_df[:,2] == ess_array
+        @test ess_df[:,3] == rhat_array
+    end
+end
```
devmotion, line 88: Nice that there are so many tests!

test/missing_tests.jl
```diff
@@ -15,23 +15,24 @@ end
...
-    describe(devnull, chn_m; sections = nothing)
-    m1, m2 = summarystats(chn_m), quantile(chn_m)
+#     # Call describe with missing values.
+#     describe(devnull, chn_m; sections = nothing)
+#     m1, m2 = summarystats(chn_m), quantile(chn_m)
 
-    @test testdiff(s1, m1)
-end
+#     @test testdiff(s1, m1)
+# end
```
devmotion, line 35: In general, I think it is better to remove stuff than to comment it (it will still be in the git history if we agree to remove it).

test/ess_tests.jl
devmotion: IIRC this PR fixes https://github.com/TuringLang/MCMCChains.jl/issues/137#issuecomment-552213216, so it would be good to add a test for it.

cpfiffer:
I think David has captured my comments (and more) so I'm ready to merge when his comments are addressed and he gives a thumbs-up.

Great work!

src/ess.jl
```diff
@@ -0,0 +1,353 @@
...
+    return ChainDataFrame("ESS", nt)
+end
+
+function ess_rhat(chains::AbstractArray{<:Union{Missing,Real},3};
+                  method::AbstractESSMethod = ESSMethod(), kwargs...)
+    return ess_rhat(method, chains; kwargs...)
+end
+
+function ess_rhat(method, chains::AbstractArray{<:Union{Missing,Real},3};
+                  maxlag::Int = 250)
```
devmotion, line 217: Just noticed that we could just define the keyword argument version
```suggestion
function ess_rhat(
    chains::AbstractArray{<:Union{Missing,Real},3};
    method::AbstractESSMethod = ESSMethod(),
    maxlag::Int = 250
)
```
since we don't dispatch on `method`. Julia will still compile optimized code for each method.

src/stats.jl
devmotion, line 262: Probably we should also allow to choose the ESS method here, in case you want the summary but not the default ESS algorithm (because it's too slow or whatever). Maybe just add an additional keyword argument - or maybe it's easier to specify the keyword arguments for `mcse` explicitly? I don't remember which arguments one can pass to that function.

src/ess.jl
```diff
@@ -0,0 +1,351 @@
...
+
+    # define cache for the computation of the autocorrelation
+    esscache = build_cache(method, samples, chain_var)
+ 
+    # define output arrays
+    ess = Vector{T}(undef, nparams)
+    rhat = Vector{T}(undef, nparams)
+
+    # for each parameter
+    for (i, chains_slice) in enumerate(eachcol(chains))
```
devmotion, line 244: `eachcol` does not exist in Julia 1.0, I guess it's best if we add `Compat` (just add the package and load it).

denainjs: ie include it in the dependencies in `Project.toml`, and write `using Compat` at the beginning of `src/MCMCChains.jl`?

devmotion: You should set the compatibility bound to 2.2.0 (that's when `eachcol` was added to Compat).

devmotion: > ie include it in the dependencies in Project.toml, and write using Compat at the beginning of src/MCMCChains.jl?

Yes, exactly.

devmotion: Oops sorry, I didn't know that `eachcol` is apparently only defined for matrices, so you have to include both Compat (for Julia 1.0) and revert this line to
```suggestion
    for (i, chains_slice) in enumerate(eachslice(chains; dims = 2))
```

devmotion: :grimacing:

src/stats.jl
denainjs: I think it's simpler to add `method` as an explicit keyword argument of `summarystats` with default value `ESSMethod()`, and to call `ess` with `method = method`?

devmotion: Sure we can do that but I was wondering what we actually pass to `mcse` (and in particular if it's less than two keyword arguments) since if we add `method` as keyword argument we specify two keyword arguments (and their defaults) for `ess`. In general, I don't like if default keyword arguments are repeated in multiple places (since one has to keep track that they don't diverge). It seems `mcse` actually takes at most one keyword argument (https://github.com/TuringLang/MCMCChains.jl/blob/dev/src/mcse.jl) and hence the following could be an alternative:
- Use dedicated method structs for `mcse` such that we can dispatch on them instead of checking symbols
- Add `size` as a parameter of the method for `:bm`
- Don't specify `maxlag` and `essmethod` in this method explicitly and just forward the remaining keyword arguments to `ess` (since no keyword arguments are required for `mcse` anymore)

denainjs: So, something like the following in `mcse.jl`

`abstract type AbstractEtype end

struct BM <: AbstractEtype 
    size :: Integer
    function BM(_size::Integer)
        size = _size
    end
end
BM() = BM(100)

struct IMSE <: AbstractEtype end

struct IPSE <: AbstractEtype end

function mcse(etype::BM, x::Vector{<:Real})
    size = etype.size
    ...
end

function mcse(::IMSE, x::Vector{<:Real})
    ...
end

function mcse(::IPSE, x::Vector{<:Real})
    ...
end`

and declaring `summarystats` in `stats.jl` as follows

`function summarystats(
    chains::Chains;
    sections = _default_sections(chains),
    append_chains::Bool = true,
    etype::AbstractEtype = IMSE(),
    kwargs... 
)`

then calling `ess_df = ess(_chains; sections = nothing, kwargs)`?

devmotion: Yep, along these lines. But I realized that it's probably best to go with your version for now, one could change it then in a separate PR.

src/MCMCChains.jl
```diff
@@ -2,9 +2,10 @@ module MCMCChains
 
 using AxisArrays
 const axes = Base.axes
-
+using AbstractFFTs
 import AbstractMCMC
-import AbstractMCMC: chainscat
+import AbstractMCMC: chainscat, AbstractChains
```
devmotion, line 7: I don't think we use `AbstractChains` anywhere? Probably it's just left from the initial PR:
```suggestion
import AbstractMCMC: chainscat
```

src/MCMCChains.jl
```diff
@@ -17,10 +18,10 @@ import Tables
 import TableTraits
 import IteratorInterfaceExtensions
 
-using LinearAlgebra: diag
+using LinearAlgebra # : diag, BlasReal, dot
```
devmotion, line 21: Would
```suggestion
using LinearAlgebra: diag, dot
```
be sufficient? I think it's nice if we can clearly state which functions we actually use from each package.

src/stats.jl
```diff
@@ -219,150 +219,30 @@ function quantile(
...
-    1 - V / (2*v)
-end
-
-"""
-    summarystats(chains[;
-                 sections, append_chains = true, maxlag = 250, etype = :bm, kwargs...])
+    function summarystats(
+        chains;
+        sections = _default_sections(chains),
+        append_chains= true,
```
devmotion, line 227: Maybe add
```suggestion
     method::AbstractESSMethod = ESSMethod(),
```
?

src/MCMCChains.jl
denainjs: I wrote
```suggestion
using LinearAlgebra : diag, BlasReal, dot
```
and used `BlasReal` instead of `LinearAlgebra.BlasReal` in `src/ess.jl`

devmotion: Ah I see :+1:

yebai:
thanks, @denainjs!

cpfiffer:
Spectacular work. I'm very happy to have this in the codebase.

PR #216, opened 2020-05-10
lower before upper in HPD

karajan9:
This should be more natural to understand if the info is printed.


cpfiffer:
Thanks, @karajan9! I concur.

PR #217, opened 2020-05-17
Fix diagnostic bug

cpfiffer:
A bug was reported in Slack where `gewekediag` was erroring out because it included names from outside the sections chosen in the chain. This fixes that and updates the testing suite to make diagnostic testing actually do something.


devmotion:
Have you checked if that is fixed in the dev branch? I know that basically all diagnostic functions are updated there in some way or another. If the bug is already fixed in the branch, we should maybe just fix the merge conflicts and get the changes into master. If not, it might still be a good idea to fix it in the dev branch, and merge it into master afterwards.

cpfiffer:
Yeah, it's already fixed in the dev branch, so we'll just overwrite this with the dev branch. But I think keep the expanded tests I wrote are worth keeping, and I'd like to have a fix out in the minor releases.

devmotion:
We should definitely add the tests, I'm just a bit worried about us accumulating even more merge conflicts.

cpfiffer:
I think, at least for these files, we can safely assume that all the diagnostic stuff is unambiguously better in the dev branch and the master can safely be overwritten.

PR #218, opened 2020-05-20
Update master

devmotion:


Project.toml
```diff
@@ -2,41 +2,54 @@ name = "MCMCChains"
...
 StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
+TableTraits = "3783bdb8-4a98-5b6b-af9a-565f29a5fe9c"
+Tables = "bd369af6-aec1-5ad0-b16a-f7cc5008161c"
 
 [compat]
+AbstractFFTs = "0.5"
 AbstractMCMC = "0.4, 0.5, 1.0"
-AxisArrays = "0.4.2"
-Distributions = "^0.21, 0.22, 0.23"
+AxisArrays = "0.3, 0.4"
```
devmotion, line 31: Just noticed that we want to keep the lower bound here:
```suggestion
AxisArrays = "0.4.2"
```

PR #225, opened 2020-06-26
Fix to heideldiag for short unconverged chains

mmattocks:
Chains under 10 steps in length currently have a delta of zero in the heideldiag while loop, which prevents loop exit if converged==false.


mmattocks:
This is a bigger problem than the commit suggests, opening issue

PR #227, opened 2020-07-02
Remove `read` and `write`

devmotion:
This PR removes the implementations of `Base.write` and `Base.read` which just (de)serialize a `Chains` object.

Copied from Slack:

> IMO write suggests that you can just save your object and load it at any time (like e.g. a CSV file), but actually in general serialization and deserialization "will not work if the reading and writing are done by different versions of Julia, or an instance of Julia with a different system image." https://docs.julialang.org/en/v1/stdlib/Serialization/#Serialization-1

I added deprecations and updated the README accordingly.


cpfiffer:
I think if we were going to merge this, we should also provide more robust saving/loading tools than just serialization, rather than just deprecating to a more simplified serialization suite. 

It's been on the docket for a while, I think, but this PR might be the impetus to provide way better saving/loading functionality.

devmotion:
Yes, I guess it would be nice to provide proper alternatives. That being said, I don't think that this PR leads to any more complicated code for users on Julia >= 1.1 - on the contrary, actually it removes an argument since it replaces `write(file, chain)` with `serialize(file, chain)` and `read(file, Chains)` with `deserialize(file)` while at the same time we are more honest about what is going on and that it might be dangerous for persistent storage (on Julia < 1.1, one has to wrap it in `open` and `read` statements since `serialize` and `deserialize` don't accept strings).

In general, it might be better to point users towards [JLSO](https://github.com/invenia/JLSO.jl) instead of using `serialize`/`deserialize` directly.

cpfiffer (approved):

README.md
```diff
@@ -298,14 +298,18 @@ corner(c::Chains, [:A, :B])
 
 ### Saving and Loading Chains
 
-Chains objects can be serialized and deserialized using `read` and `write`.
+Like any Julia object, a `Chains` object can be saved using `Serialization.serialize`
+and loaded back by `Serialization.deserialize` as identical as possible.
+Note, however, that in general
+[this process will not work if the reading and writing are done by different versions of Julia, or an instance of Julia with a different system image.](https://docs.julialang.org/en/v1/stdlib/Serialization/#Serialization-1)
```
devmotion, line 304:
```suggestion
[this process will not work if the reading and writing are done by different versions of Julia, or an instance of Julia with a different system image](https://docs.julialang.org/en/v1/stdlib/Serialization/#Serialization-1).
You might want to consider [JLSO](https://github.com/invenia/JLSO.jl) for saving metadata
such as the Julia version and the versions of all packages installed as well.
```

PR #229, opened 2020-07-29
Fix ESS

devmotion:
This PR fixes a bug in the estimation of the effective sample size that was brought up on Slack by @yiyuezhuo (I guess?).

The current implementation combines the autocorrelation estimates of each chain incorrectly. I assume I based the current implementation (partly) on https://mc-stan.org/docs/2_18/reference-manual/effective-sample-size-section.html (since it was bookmarked on my computer) which, however, is an old version and contains an error in the equation for the combined autocorrelations in section 15.4.2 (it is fixed in the latest version).

I extended the documentation to explain the algorithms more clearly as well.

Unfortunately the bug was not caught by any of the tests, I'm not sure why that's the case. Maybe we should include a CSV file with the data of the MWE on Slack as a test?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -3,7 +3,7 @@ uuid = "c7f686f2-ff18-58e9-bc7b-31028e88f75d"
 keywords = ["markov chain monte carlo", "probablistic programming"]
 license = "MIT"
 desc = "Chain types and utility functions for MCMC simulations."
-version = "4.0.1"
+version = "4.0.2"
 
 [deps]
 AbstractFFTs = "621f4979-c628-5d54-868e-fcf4e3e8185c"
--- a/src/ess.jl
+++ b/src/ess.jl
@@ -100,9 +100,9 @@ function build_cache(::BDAESSMethod, samples::Matrix, var::Vector)
     return BDAESSCache(samples, var, mean(var))
 end
 
-update_cache!(cache::ESSCache) = nothing
+update!(cache::ESSCache) = nothing
 
-function update_cache!(cache::FFTESSCache)
+function update!(cache::FFTESSCache)
     # copy samples and add zero padding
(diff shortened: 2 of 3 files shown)


yiyuezhuo:
Maybe there're many wrong equations in Stan's document... I just found [one](https://mc-stan.org/docs/2_23/reference-manual/correlation-matrix-transform-section.html) yesterday as well...

denainjs:
If I understand correctly, the main problem is that `ess_rhat` failed to take into account the `s_m^2` in equation (10) [here](https://arxiv.org/pdf/1903.08008.pdf). 

If so, perhaps the reason the tests passed despite the first problem is the following:

- [this](https://github.com/TuringLang/MCMCChains.jl/blob/53c87c270369d5ec0248d49fb0ef1e665af4de3e/test/ess_tests.jl#L31) testset is pretty much the only one in `ess_tests.jl` that checks that the ess estimates return values that make sense (the other testsets look at even more degenerate settings, or check that shapes/types work out).

- it uses samples from iid standard normal variables to fill the `Chain`.

- the `s_m^2` in equation (10) are estimators of the variance of this distribution, hence close to 1.

- therefore omitting them didn't change the result much, in this particular case.

When I multiplied the standard normal samples by 50, the tests failed.

devmotion:
Great idea, I like this a lot better than including some random data.

> If I understand correctly, the main problem is that ess_rhat failed to take into account the s_m^2 in equation (10) here.

Yes, exactly.

> If so, perhaps the reason the tests passed despite this is the following:

Yep, I suspected something along these lines.

devmotion:
I added the scaled tests which failed due to the previous bug.

devmotion:
> I'm doing my tests today

Good luck! I'll take care of it.

PR #234, opened 2020-08-19
Change convert(::Type{Array}, ::Vector{<:ChainDataFrame}) to the

### LINKED ISSUE #233: Avoid subtle piracy in convert or handle the pirated cases
The method https://github.com/TuringLang/MCMCChains.jl/blob/78448b71aac21437900ac59f689b46e23c5d0cd6/src/summarize.jl#L118-L122 is actually piracy since `Union{}<:ChainDataFrame` and `Union` isn't defined here. It's unclear to me why you need the method (why isn't it a noop), but if it's really needed then you'd have to handle the `Union{}` case since currently
```julia
julia> convert(Array, Vector{Union{}}())
Union{}[]

julia> using MCMCChains

julia> convert(Array, Vector{Union{}}())
ERROR: ArgumentError: reducing over an empty collection is not allowed
Stacktrace:
 [1] _empty_reduce_error() at ./reduce.jl:299
 [2] mapreduce_empty(::Function, ::Function, ::Type{T} where T) at ./reduce.jl:342
 [3] reduce_empty(::Base.MappingRF{MCMCChains.var"#105#107",MCMCChains.var"#106#108"}, ::Type{Union{}}) at ./reduce.jl:329
 [4] reduce_empty_iter at ./reduce.jl:355 [inlined]
 [5] mapreduce_empty_iter(::Function, ::Function, ::Array{Union{},1}, ::Base.HasEltype) at ./reduce.jl:351
 [6] _mapreduce(::MCMCChains.var"#105#107", ::MCMCChains.var"#106#108", ::IndexLinear, ::Array{Union{},1}) at ./reduce.jl:400
 [7] _mapreduce_dim(::Function, ::Function, ::NamedTuple{(),Tuple{}}, ::Array{Union{},1}, ::Colon) at ./reducedim.jl:318
 [8] #mapreduce#620 at ./reducedim.jl:310 [inlined]
 [9] mapreduce(::Function, ::Function, ::Array{Union{},1}) at ./reducedim.jl:310
 [10] convert(::Type{Array}, ::Array{Union{},1}) at /Users/andreasnoack/.julia/packages/MCMCChains/0Cwci/src/summarize.jl:119
 [11] top-level scope at REPL[3]:1
```

andreasnoack:
internal _to_matrix function since the source is already an Array
so the operation should really be a noop. Furthermore, the implied
type piracy causes issues in other packages.

Closes #233

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -3,7 +3,7 @@ uuid = "c7f686f2-ff18-58e9-bc7b-31028e88f75d"
 keywords = ["markov chain monte carlo", "probablistic programming"]
 license = "MIT"
 desc = "Chain types and utility functions for MCMC simulations."
-version = "4.0.2"
+version = "4.0.3"
 
 [deps]
 AbstractFFTs = "621f4979-c628-5d54-868e-fcf4e3e8185c"
--- a/src/summarize.jl
+++ b/src/summarize.jl
@@ -106,7 +106,7 @@ end
 function Base.convert(::Type{Array}, c::C) where C<:ChainDataFrame
     T = promote_eltype_namedtuple_tail(c.nt)
     arr = Array{T, 2}(undef, c.nrows, c.ncols - 1)
-    
+
     for (i, k) in enumerate(Iterators.drop(keys(c.nt), 1))
         arr[:, i] = c.nt[k]
     end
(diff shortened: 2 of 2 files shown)


andreasnoack:
What's on the current master? Could we make a fast bugfix release with this to resolve https://github.com/JuliaData/SentinelArrays.jl/issues/31?

devmotion:
Just updated the README yesterday but otherwise no changes since the last release, so a bugfix release is fine :+1:

devmotion (approved):
LGTM

andreasnoack:
Actually. Even though I don't think the original method makes sense (it should be a noop), it's a public function that people might rely on so let me change it to the one you original proposed to minimize the risk.

devmotion (approved):
Thanks!

PR #238, opened 2020-09-03
Adding implementation of Rstar statistic

trappmartin:
This PR adds an implementation of the R* statistic described in https://arxiv.org/pdf/2003.07900.pdf.

Todo: 
- [x] ~~Add additional tests.~~
- [x] ~~My implementation of Algo. 2 (generating multiple draws of R*) seems to result in rather low values on the unit-test set. Have to inspect if there is a subtle bug.~~
- [x] ~~Add documentation on the README.md~~
- [x] ~~Replace XGBoost with MLJ interface if possible~~

Thanks @ben18785 and @avehtari for the nice paper.


devmotion:
I think it might make sense to implement this statistic in a separate package but not in MCMCChains. Requires is pretty bad and actually not needed anymore in MCMCChains, we removed all optional package dependencies to decrease long compilation and loading times (see https://github.com/TuringLang/MCMCChains.jl/pull/239). It would also make it easier to play around with and support different classifiers and other packages in addition to XGBoost. Some additional high-level comments, without having checked the implementation in detail:
- I think it would be good to implement statistics for general array-type inputs instead of `Chains`, and only add a wrapper/dispatch for `Chains` if needed. We already changed this for a bunch of statistics in MCMCChains since then the implementation of the statistics can be focused on the actual algorithm and can be reused for other chain types (or just arrays). Moreover, it usually simplified the code and made working with `Chains` more efficient (since, e.g., the relevant information can be extracted once in the beginning instead of subsetting the chain multiple times).
- Following standard Julia conventions, one might want to name the function `rstar` instead of `Rstar`.
- As mentioned above, I think it would be nice to write the implementation in a generic way that supports different classifiers/classifier packages.

trappmartin:
Good points, I’ll think about it. However, I’m sceptical it’s worth implementing in a separate repo as it’s literally only a few lines. I agree that it would be good to do it more general, I’m personally not a fan of boosting. Maybe we can use the interfaces of MLJ for that.

src/rstar.jl
```diff
@@ -1,7 +1,8 @@
-
+import XGBoost
```
devmotion, line 1: Requires requires you to specify optional dependencies with a dot (the problem is that tests don't seem to cover this issue but users will get a warning otherwise):
```suggestion
using .XGBoost
```

trappmartin:
Ok, I agree that it would make sense to have an external repository containing more advanced diagnostics. Let's keep this PR open for now and merge the code into the new repo once we know what goes into the repo.

ben18785:
@trappmartin and everyone else, thanks for picking this up and glad that you are considering implementing R*. If you have any questions, let me know: @trappmartin, let me know if you continue to have issues with Algo 2 and I can cross check with mine if you give me details of the test being done.

cpfiffer:
For now, let's just stick R* in a new package (AdvancedDiagnostics?) and get it registered. Excellent work so far!

yebai:
It feels slight overkill to register and maintain a new package for this single feature. Maybe consider adding this to a ’src/contrib’ folder? 

I appreciate the idea of keeping loading time minimal, but that’s like a language level optimisation task. Therefore, reducing loading time by splitting packages sounds a bit counter-productive IMO.

src/rstar.jl
```diff
@@ -0,0 +1,93 @@
...
+
+    N = length(y)
+
+    # randomly sub-select training and testing set
+    Ntrain = round(Int, N*subset)
+    Ntest = N - Ntrain
+
+    ids = Random.shuffle(collect(1:N))
+    train_ids = ids[1:Ntrain]
+    test_ids = ids[(Ntrain+1):end]
```
devmotion, line 48:
```suggestion
    ids = Random.randperm(N)
    train_ids = view(ids, 1:Ntrain)
    test_ids = view(ids, (Ntrain+1):N)
```
Additionally, it would be nice to make a `rng::AbstractRNG` the first argument to `rstar` and use the provided random number generator here. Additionally, maybe a (additional) function that takes explicit training and test data instead of splitting the provided data might be useful.

src/rstar.jl
```diff
@@ -0,0 +1,93 @@
...
+    @assert length(test_ids) == Ntest
+
+    # use predicted probabilities?
+    mode = iterations > 1 ? "multi:softprob" : "multi:softmax"
+
+    @info "Training classifier"
+    # train classifier using XGBoost
+    classif = XGBoost.xgboost(x[train_ids,:], niter; label = y[train_ids], 
+                      objective = mode, num_class = nchains,
+                      xgboostparams...)
```
devmotion, line 59: It would be nice if one could avoid these explicit calls to XGBoost and use a more general interface.

src/rstar.jl
```diff
@@ -0,0 +1,93 @@
...
+
+    @info "Computing R* statistics"
```
devmotion, line 61: Maybe rather
```suggestion
    @debug "Computing R* statistics"
```
if we want to keep it?

src/rstar.jl
```diff
@@ -0,0 +1,93 @@
...
+    Rstats = ones(iterations) * Inf
```
devmotion, line 62:
```suggestion
    Rstats = fill(Inf, iterations)
```

src/rstar.jl
```diff
@@ -0,0 +1,93 @@
...
+    for i in 1:iterations
+        # predict labels for "test" data
+        p = XGBoost.predict(classif, x[test_ids,:])
```
devmotion, line 65: Is `XGBoost.predict` different from the general `StatsBase.predict`?
```suggestion
        p = predict(classif, x[test_ids,:])
```

devmotion: It would also be good to compute `xtest = x[test_ids, :]` and `ytest = y[test_ids]` outside of the loop to avoid allocating new arrays in every iteration.

src/rstar.jl
```diff
@@ -0,0 +1,93 @@
...
+
+        pred = if length(p) == Ntest*nchains
+            probs = reshape(p, Ntest, nchains)
+            map(s -> rand(Categorical(s / sum(s))), eachrow(probs))
```
devmotion, line 69: Why is this branch needed? It seems it was not covered by the tests BTW.

devmotion: It would be good if we could specify a `rng` here as well. Moreover, do we have to iterate over rows here, column-based indexing would be faster (I guess that's related to the fact that samples in `x` are stored per row)? One might also want to scale the matrix in-place:
```suggestion
        pred = if length(p) == Ntest*nchains
            probs = reshape(p, Ntest, nchains)
            probs ./= sum(probs; dims=2)
            map(s -> rand(Categorical(s)), eachrow(probs))
```

src/rstar.jl
```diff
@@ -0,0 +1,93 @@
...
+        else
+            p
+        end
```
devmotion, line 72: These conditional branches might lead to problems with type inference, would be good to check if the compiler can infer the type of `pred`.

src/rstar.jl
```diff
@@ -0,0 +1,93 @@
...
+
+        # compute statistic
+        a = mean(pred .== y[test_ids])
```
devmotion, line 75: As mentioned above, one should define `ytest = y[test_ids]` outside of the loop. In general, one could avoid allocations here by writing
```suggestion
        a = mean(i -> pred[i] == y[i], test_ids)
```
or
```suggestion
        a = mean(((p, y),) -> p == y, zip(pred, ytest))
```

src/rstar.jl
```diff
@@ -0,0 +1,93 @@
...
+        Rstats[i] = nchains*a
+    end
+
+    return Rstats
+end
+
+function rstar(chn::Chains, iterations; kwargs...)
+    nchains = length(chains(chn))
```
devmotion, line 83:
```suggestion
    nchains = size(chn, 3)
```

src/rstar.jl
```diff
@@ -0,0 +1,93 @@
...
+    @assert nchains > 1
```
devmotion, line 84: Again I guess it would be better to throw an actual error or remove this check.

src/rstar.jl
```diff
@@ -0,0 +1,93 @@
...
+
+    # collect data
+    x = mapreduce(c -> Array(chn[:,:,c]), vcat, chains(chn))
```
devmotion, line 87: Isn't this just
```suggestion
    x = Array(chn)
```
?

src/rstar.jl
```diff
@@ -0,0 +1,93 @@
...
+    y = mapreduce(c -> ones(Int, length(chn))*c, vcat, chains(chn)) .- 1
```
devmotion, line 88: Is it actually needed to normalize the classes to 0, ..., nchains-1? I would assume that classifiers should be agnostic to the choice of labels (in particular since I'm not completely sure if `chains` is guaranteed to be normalized to 1, ..., nchains).
```suggestion
    nsamples = size(chn, 1)
    y = repeat(chains(chn); inner = nsamples)
```

devmotion (commented):
I had a closer look at the implementation and added some comments.

src/rstar.jl
trappmartin: Nice one, thanks.

src/rstar.jl
trappmartin: Yes, I’ll have a look if this could be done with the MLJ interface. I just didn’t have the time atm.

src/rstar.jl
trappmartin: Ah, I didn’t know that. Nice

src/rstar.jl
trappmartin: XGBoost is not and requires the class labels to start with zero. The MLJ interface might be.

devmotion: In any case, I guess it would be simpler to use `repeat` here :slightly_smiling_face:

devmotion:
> I appreciate the idea of keeping loading time minimal, but that’s like a language level optimisation task. Therefore, reducing loading time by splitting packages sounds a bit counter-productive IMO.

The problem reported in https://github.com/TuringLang/Turing.jl/issues/1400 shows that providing optional features using Requires not only increases loading times but causes precompilation problems as well. Definitely, the problem of optional/conditional dependencies has to be addressed by Pkg on a global level. However, as long as it is not solved, IMO creating small packages is the more stable approach than adding optional dependencies with Requires. There are also multiple examples of small packages in the Julia ecosystem (e.g., https://github.com/SciML/PoissonRandom.jl, https://github.com/SciML/IfElse.jl, https://github.com/SciML/RuntimeGeneratedFunctions.jl, https://github.com/SciML/MuladdMacro.jl), so it is not uncommon to publish packages with only a handful of functions or one source file.

In this case, one could also bundle the Rstar statistic together with other more advanced diagnostics in MCMCChains in a separate package. IMO one should be careful to not add too many dependencies to MCMCChains (as mentioned, we had problems with really massive package loading times before) but actually I'd rather take a dependency on XGBoost than Requires here. (Though I still think it would be nice if one could provide a general implementation that works with different classification algorithms - maybe by using an MLJ interface, as @trappmartin suggested above.)

trappmartin:
No worries, I’ll have a look at the MLJ interface. I just didn’t have the time so far.

devmotion:
No worries, it was just intended as a general comment regarding optional dependencies and small packages :slightly_smiling_face:

trappmartin:
OK, so this PR is now independent of XGBoost and uses the minimal interface provided in `MLJModelInterface`. Thus, the user can effectively use any MLJ supported classifier to run the statistic on inference results.

Here is an example using XGBoost. Note that this requires the user to have MLJ, MLJModels and XGBoost installed.

```julia
julia> using MLJ, MLJModels
julia> using Turing, MCMCChains

julia> # define some Turing model
julia> @model function high_dim_gauss(D)
    m ~ filldist(Normal(0, 1), D)
end

julia> # instantiation
julia> model = high_dim_gauss(50)

julia> # sample with different settings
julia> chn1 = sample(model, NUTS(1, 0.8), MCMCDistributed(), 11, 4); # generate 10 samples (adaptation 1)
julia> chn2 = sample(model, NUTS(500, 0.8), MCMCDistributed(), 2000, 4); # generate 1500 samples (adaptation 500)

julia> # load classifier that should be used for the computation
julia> classif = @load XGBoostClassifier

julia> # estimate 10 samples of the statistic (10 is default).
julia> R1 = rstar(chn1, classif)
julia> R2 = rstar(chn2, classif)

julia> # some plotting
julia> using UnicodePlots
julia> boxplot(["N=10", "N=1500"], [R1, R2])
          ┌                                        ┐
                                   ╷        ┌─────╷
     N=10                          ├────────┤     ┤
                                   ╵        └─────╵
           ╷┐╷
   N=1500  ├├┤
           ╵┘╵
          └                                        ┘
          1                  2.5                   4

julia> histogram(R2)
                ┌                                        ┐
   [1.05, 1.1 ) ┤▇▇▇▇▇ 1
   [1.1 , 1.15) ┤▇▇▇▇▇ 1
   [1.15, 1.2 ) ┤▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ 7
   [1.2 , 1.25) ┤▇▇▇▇▇ 1
                └                                        ┘
                                Frequency
```

src/rstar.jl
```diff
@@ -0,0 +1,125 @@
+# Data structure required to store samples for MLJ.
+struct RStarTable{T<:AbstractFloat}
```
devmotion, line 2: Why is this needed? Can't you just provide a [`MatrixTable`](https://github.com/JuliaData/Tables.jl/blob/master/src/matrix.jl) to the MLJ model? I.e., users provide the data as matrices to `rstar` and you call `table = Tables.table(data)` internally?

src/rstar.jl
```diff
@@ -0,0 +1,125 @@
...
+        # compute statistic
+        a = mean(((p,y),) -> p == y, zip(pred, ytest))
+        Rstats[i] = K*a
+    end
+
+    return Rstats
+end
+
+function rstar(chn::Chains, model::MLJModelInterface.Supervised; kwargs...)
+    return rstar(Random.default_rng(), chn, model; kwargs...)
```
devmotion, line 113:
```suggestion
    return rstar(Random.GLOBAL_RNG, chn, model; kwargs...)
```

src/rstar.jl
```diff
@@ -0,0 +1,125 @@
...
+        p = MLJModelInterface.predict(classif, fitresult, xtest)
+        pred = similar(ytest)
+
+        if classif isa Deterministic
+            pred[:] = p
+        elseif classif isa Probabilistic
+            pred[:] = get.(rand.(rng, p))
+        else
+            throw(ErrorException("Unknown type of classifier"))
+        end
```
devmotion, line 102: I guess it would be easier to not preallocate `pred` but compute it directly:
```suggestion
        if classif isa Deterministic
            pred = MLJModelInterface.predict(classif, fitresult, xtest)
        elseif classif isa Probabilistic
            pred = get.(rand.(rng, MLJModelInterface.predict(classif, fitresult, xtest)))
        end
```
Also the type of classifier can be checked outside of the loop.

src/rstar.jl
```diff
@@ -0,0 +1,125 @@
...
+    Ntest = N - Ntrain
+
+    ids = Random.randperm(rng, N)
+    train_ids = view(ids, 1:Ntrain)
+    test_ids = view(ids, (Ntrain+1):N)
+
+    # train classifier using XGBoost
+    fitresult, _ = MLJModelInterface.fit(classif, verbosity, x[train_ids], MLJModelInterface.categorical(y[train_ids]))
+
+    Rstats = fill(Inf, iterations)
```
devmotion, line 86: I suggest to not preallocate `Rstats` but rather move the body of the for loop into a separate function, with the changes suggested below. Then you can also dispatch on `classif` instead of hardcoding this check. `Rstats` could you then create e.g. with a list comprehension:
```julia
    Rstats = [K * rstar_predict(rng, classif, fitresult, xtest, ytest) for _ in 1:iterations]
```

trappmartin:
@devmotion thanks for those useful comments!

src/rstar.jl
trappmartin: Good idea!

src/rstar.jl
trappmartin: Great, I wasn't aware of MatrixTable.

trappmartin:
This PR should be ready to merge once all tests run through.

src/MCMCChains.jl
```diff
@@ -12,7 +12,7 @@ using SpecialFunctions
 using Formatting
 import StatsBase: autocov, counts, sem, AbstractWeights,
     autocor, describe, quantile, sample, summarystats, cov
-
+using MLJModelInterface
```
devmotion, line 15: Since MLJModelInterface is specified explicitly in almost all (apart from two?) places, it might make sense to only import it:
```suggestion
import MLJModelInterface
```
IMO this also helps to quickly figure out which methods and types are actually from MLJModelInterace.

src/rstar.jl
```diff
@@ -0,0 +1,91 @@
+"""
+    rstar(chains::Chains, model::MLJModelInterface.Supervised; subset = 0.8, iterations = 10, verbosity = 0)
+    rstar(rng::Random.AbstractRNG, chains::Chains, model::MLJModelInterface.Supervised; subset = 0.8, iterations = 10, verbosity = 0)
+    rstar(rng::Random.AbstractRNG, model::MLJModelInterface.Supervised, x::AbstractMatrix, y::AbstractVector; subset = 0.8, iterations = 10, verbosity = 0)
+
+Compute the R* statistic for convergence diagnostic of MCMC. This implementation is an adaption of Algorithm 1 & 2, described in [Lambert & Vehtari]. Note that the correctness of the statistic depends on the convergence of the classifier used internally in the statistic. You can track if the training of the classifier converged by inspection of the printed RMSE values from the XGBoost backend. To adjust the number of iterations used to train the classifier set `niter` accordingly.
```
devmotion, line 6: It seems a bit strange that the classifier has to be specified first when provide data as matrix and vector, but after the chains argument in the other case. Might be good to decide for one consistent form. Also might be useful to allow to call `rstar` with matrix+vector argument without specifying an RNG.

Maybe also simplify the docstring a bit according to the [docs](https://docs.julialang.org/en/v1/manual/documentation/#Documentation), e.g., it is not necessary to specify the keyword arguments in the signature since you list them in a separate section. I think it's also not needed to add all type annotations (in particular for the RNG). E.g., something like:
```suggestion
    rstar([rng ,]model, chains::Chains; kwargs...)
    rstar([rng ,]model, x::AbstractMatrix, y::AbstractVector; kwargs...)

Compute the R* convergence diagnostic of MCMC.

This implementation is an adaption of Algorithm 1 & 2, described in [Lambert & Vehtari]. Note that the correctness of the statistic depends on the convergence of the classifier used internally in the statistic. You can track if the training of the classifier converged by inspection of the printed RMSE values from the XGBoost backend. To adjust the number of iterations used to train the classifier set `niter` accordingly.
```
Maybe even more descriptive to use `classifier` instead of `model` in the docs?

src/rstar.jl
```diff
@@ -0,0 +1,91 @@
...
+
+# Keyword Arguments
+* `subset = 0.8` ... Subset used to train the classifier, i.e. 0.8 implies 80% of the samples are used.
+* `iterations = 10` ... Number of iterations used to estimate the statistic. If the classifier is not probabilistic, i.e. does not return class probabilities, it is advisable to use a value of one.
+* `verbosity = 0` ... Verbosity level used during fitting of the classifier.
+
+# Usage
+
+```julia-repl
```
devmotion, line 15: I think `julia-repl` should only be used if the code is actually specified in the Julia REPL format, i.e., as
```julia
julia> ...
```
Maybe just change it to `julia`?

devmotion (commented):
Looks good, just left some additional comments/questions. I like that we can make use of all the models implementing the MLJ interface.

src/MCMCChains.jl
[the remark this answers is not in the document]
trappmartin: Perfect!

src/rstar.jl
trappmartin: Yes, good points. Thanks.

src/rstar.jl
```diff
@@ -0,0 +1,92 @@
...
+
+## References:
+[Lambert & Vehtari] Ben Lambert and Aki Vehtari. "R∗: A robust MCMC convergence diagnostic with uncertainty using gradient-boostined machines." Arxiv 2020.
+"""
+function rstar(rng::Random.AbstractRNG, classif::MLJModelInterface.Supervised, x::AbstractMatrix, y::AbstractVector{Int}; iterations = 10, subset = 0.8, verbosity = 0)
+
+    size(x,1) != length(y) && throw(DimensionMismatch())
+    iterations >= 1 && ArgumentError("Number of iterations has to be positive!")
+
+    if iterations > 1 && classif isa Deterministic
```
devmotion, line 42:
```suggestion
    if iterations > 1 && classif isa MLJModelInterface.Deterministic
```

src/rstar.jl
```diff
@@ -0,0 +1,92 @@
...
+        @warn("Classifier is not a probabilistic classifier but number of iterations is > 1.")
+    elseif iterations == 1 && classif isa Probabilistic
```
devmotion, line 44:
```suggestion
    elseif iterations == 1 && classif isa MLJModelInterface.Probabilistic
```

src/rstar.jl
devmotion: I reopened this conversation since it seems the latest commits did not address the comments about the method inconsistency and the version without RNG.

Project.toml
```diff
@@ -14,6 +14,7 @@ Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
 Formatting = "59287772-0a20-5a39-b81b-1366585eb4c0"
 IteratorInterfaceExtensions = "82899510-4779-5014-852e-03e436cf321d"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+MLJModelInterface = "e80e1ace-859a-464e-9ed9-23947d8ae3ea"
```
devmotion, line 17: Can you add a MLJModelInterface to the `[compat]` section as well? Otherwise the merge bot in the Julia registry will complain.

Additionally, it would be good to bump the version number to `4.2.0` as well when finishing up the PR :slightly_smiling_face:

src/rstar.jl
trappmartin: It’s already resolved locally but I didn’t have time to push. Will do tomorrow.

Project.toml
trappmartin: Sounds good.

Project.toml
```diff
@@ -33,6 +34,7 @@ Compat = "2.2, 3"
 Distributions = "0.21, 0.22, 0.23"
 Formatting = "0.4"
 IteratorInterfaceExtensions = "0.1.1, 1"
+MLJModelInterface = "~0.3.5"
```
devmotion, line 37: For pre-1.0 versions, this is actually [the same as the standard caret specifier which can be omitted](https://julialang.github.io/Pkg.jl/v1/compatibility/#Caret-specifiers-1):
```suggestion
MLJModelInterface = "0.3.5"
```

trappmartin: OK.

devmotion (approved):
LGTM, great work!

cpfiffer:
Not it I beat you to it!

PR #239, opened 2020-09-03
Remove unused Requires dependency

devmotion:
Requires is not actually used anymore, so let's improve loading and compilation
times even more and remove it completely.


cpfiffer (approved):
Looks good, thanks for catching this one.

PR #241, opened 2020-09-29
Fix deprecated `read` and `write`

devmotion:
Fixes a bug introducing when deprecating `read` and `write`. Instead of overloading Base.read and Base.write we accidentally added new MCMCChains.read and MCMCChains.write methods.


yebai (approved):
Thanks, @devmotion -- Looks good to me!

PR #245, opened 2020-10-11
Fix ChainDataFrame indexing with Colon

### LINKED ISSUE #244: ChainDataFrame slice ignores symbol key
I suppose the following is not intentional?

```julia
julia> e[Symbol("w[1]"), :]
ESS
  parameters     ess   r_hat
  ──────────  ──────  ──────
        w[1]  4.4639  1.0522
        w[2]  4.4639  1.0522
        z[1]     NaN     NaN
        z[2]     NaN     NaN
        z[3]     NaN     NaN
        z[4]     NaN     NaN
        z[5]     NaN     NaN
        z[6]     NaN     NaN
        z[7]     NaN     NaN
        z[8]     NaN     NaN
        z[9]     NaN     NaN
       z[10]     NaN     NaN
        μ[1]  4.9389  1.0134
        μ[2]  4.3662  1.0619


julia> e[Symbol("w[1]")]
ESS
  parameters     ess   r_hat
  ──────────  ──────  ──────
        w[1]  4.4639  1.0522
```

Looks like [this line][1] accidentially ignores the symbol key.

[1]: https://github.com/TuringLang/MCMCChains.jl/blob/fa6757ad3044abf1dc8f8cee3f0bd0a690899cf8/src/summarize.jl#L42

cpfiffer:
Fixes #244. The buggy function was actually identical to the one below it, so I just mapped them together.

### DIFF
--- a/test/summarize_tests.jl
+++ b/test/summarize_tests.jl
@@ -13,7 +13,11 @@ using Statistics: std
 
     @test 0.48 < parm_df[:a, :mean][1] < 0.52
     @test names(parm_df) == [:parameters, :mean, :std, :naive_se, :mcse, :ess, :rhat]
-    @test parm_df[:a, :] == parm_df[:a]
+
+    # Indexing tests
+    @test convert(Array, parm_df[:a, :]) == convert(Array, parm_df[:a])
+    @test parm_df[:a, :][:,:parameters] == :a
+    @test parm_df[[:a, :b], :][:,:parameters] == [:a, :b]
 
     all_sections_df = summarize(chns, sections=[:parameters, :internals])
     @test all_sections_df[:,:parameters] == [:a, :b, :c, :d, :e, :f, :g, :h]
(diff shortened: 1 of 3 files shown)


devmotion:
Can you add a simple test and bump the version number for a bugfix release?

test/summarize_tests.jl
```diff
@@ -13,6 +13,7 @@ using Statistics: std
 
     @test 0.48 < parm_df[:a, :mean][1] < 0.52
     @test names(parm_df) == [:parameters, :mean, :std, :naive_se, :mcse, :ess, :rhat]
+    @test parm_df[:a, :] == parm_df[:a]
```
devmotion, line 16: Maybe check also explicitly that it does not return other parameters as well? Otherwise it could be that both implementations are incorrect, couldn't it?

cpfiffer: We check the other implementations down below, like this one, for example:

```julia
@test three_parms_df_2[:, :parameters] == [:a, :b, :g]
```

cpfiffer: No, nevermind, that indexing is done at chain level. Give me a sec to add that too.

devmotion (approved):
Thanks, looks good to me.

Issue #250, opened 2020-11-16
Is borderless still supported in PrettyTables (summarize.jl line 31)


goedman:
This is in relation to https://github.com/TuringLang/MCMCChains.jl/pull/247#issue-516411334.

I'm getting the following error:
```
Chains MCMC chain (1000×13×4 Array{Float64,3}):

Iterations        = 1:1000
Thinning interval = 1
Chains            = 1, 2, 3, 4
Samples per chain = 1000
parameters        = θ
internals         = acceptance_rate, hamiltonian_energy, hamiltonian_energy_error, is_accept, log_density, lp, max_hamiltonian_energy_error, n_steps, nom_step_size, numerical_error, step_size, tree_depth

Summary Statistics
ERROR: LoadError: UndefVarError: borderless not defined
Stacktrace:
 [1] show(::IOContext{REPL.Terminals.TTYTerminal}, ::MIME{Symbol("text/plain")}, ::ChainDataFrame{NamedTuple{(:parameters, :mean, :std, :naive_se, :mcse, :ess, :rhat),Tuple{Array{Symbol,1},Array{Float64,1},Array{Float64,1},Array{Float64,1},Array{Float64,1},Array{Float64,1},Array{Float64,1}}}}) at /Users/rob/.julia/packages/MCMCChains/PMgw9/src/summarize.jl:31
 [2] show(::IOContext{REPL.Terminals.TTYTerminal}, ::MIME{Symbol("text/plain")}, ::Chains{Float64,AxisArrays.AxisArray{Float64,3,Array{Float64,3},Tuple{AxisArrays.Axis{:iter,StepRange{Int64,Int64}},AxisArrays.Axis{:var,Array{Symbol,1}},AxisArrays.Axis{:chain,UnitRange{Int64}}}},Missing,NamedTuple{(:parameters, :internals),Tuple{Array{Symbol,1},Array{Symbol,1}}},NamedTuple{(),Tuple{}}}) at /Users/rob/.julia/packages/MCMCChains/PMgw9/src/chains.jl:251
 [3] display(::REPL.REPLDisplay, ::MIME{Symbol("text/plain")}, ::Any) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.5/REPL/src/REPL.jl:214
[...]

PR #251, opened 2020-11-16
Use `tf_borderless` for PrettyTables >= 0.10

devmotion:
Fixes https://github.com/TuringLang/MCMCChains.jl/issues/250.


goedman:
Definitely seems to work on 1.5 (and 1.6-DEV with Formatter.jl patch).

cpfiffer (approved):

PR #254, opened 2020-11-25
Switch from Travis to Github Actions

devmotion:


.github/workflows/CompatHelper.yml
```diff
@@ -13,4 +12,5 @@ jobs:
       - name: CompatHelper.main()
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          COMPATHELPER_PRIV: ${{ secrets.COMPATHELPER_PRIV }}
```
devmotion, line 15: This requires a new `COMPATHELPER_PRIV` secret (see https://github.com/JuliaRegistries/CompatHelper.jl#12-set-up-the-ssh-deploy-key-optional).

devmotion:
Oosh, I'm surprised we've never seen these test failures before since the failing test is just stupid. We test the keys in the resulting NamedTuple (and their specific order!) in https://github.com/TuringLang/MCMCChains.jl/blob/58fbeadf189d246b62675a4afe3551f1ee0ad668/test/get_tests.jl#L34 but actually the NamedTuple is constructed from a dictionary that is initialized in https://github.com/TuringLang/MCMCChains.jl/blob/58fbeadf189d246b62675a4afe3551f1ee0ad668/src/chains.jl#L157
I assume we were just lucky and on all tested platforms the iteration over the dictionary was performed in the desired order but this is not guaranteed. So the question is: is this a bug and should be changed or should we just fix the test? @cpfiffer

cpfiffer:
Yeah, that's just a stupid test. The one quick fix is just to sort the left side of that equality. #255 should fix that.

devmotion:
XGBoost seems to be broken on 32bit systems (https://github.com/dmlc/XGBoost.jl/issues/92), so I disabled it.

cpfiffer (approved):

.github/workflows/CompatHelper.yml
devmotion: Can you fix this @cpfiffer? Or have you already added one?

cpfiffer: Nope, crap, I missed that. I'll add one now.

cpfiffer: Added.

devmotion: Thank you!

Issue #258, opened 2021-01-12
Function to check HMC diagnostics


treigerm:
Stan has a nice automated way to check whether any of the HMC diagnostics hint at problems in the inference. In CmdStan this can be done using [diagnose](https://mc-stan.org/docs/2_25/cmdstan-guide/diagnose.html). The source code for it can found at https://github.com/stan-dev/cmdstan/blob/develop/src/cmdstan/diagnose.cpp. PyStan has a similar [functionality](https://github.com/stan-dev/pystan/blob/develop/pystan/diagnostics.py).

I have implemented a rough version of some of these checks for my own use but it would be nice if they would already come included in MCMCChains. I will open a pull request with the code I currently have and then maybe someone can judge whether it would be appropriate to include in MCMCChains?

treigerm:
Closing this in favour of a new [issue in Turing](https://github.com/TuringLang/Turing.jl/issues/1524).

PR #259, opened 2021-01-12
Initial draft for checking diagnostics

treigerm:
This is a pull request corresponding to #258.

If you think this is the right place for these checks then I would still have a couple of questions:
- Should I just copy the behaviour of the Stan diagnostic tools and then reference the Stan docs somewhere?
- How should logging be done? With `@info` or something else?
- Should there be different verbosity levels?

Once those questions are settled I will implement checks for the other diagnostics i.e. treedepth and E-BFMI value and also add some tests.

Please let me know any other things that could be improved!

### DIFF
--- a/src/MCMCChains.jl
+++ b/src/MCMCChains.jl
@@ -38,6 +38,8 @@ export hpd, ess
 
 export rstar
 
+export check_diagnostics, check_div, check_n_eff, check_rhat
+
 export ESSMethod, FFTESSMethod, BDAESSMethod
 
 """
@@ -76,6 +78,7 @@ include("modelstats.jl")
 include("plot.jl")
 include("tables.jl")
 include("rstar.jl")
+include("check_diagnostics.jl")
 
 # deprecations
 # TODO: Remove dependency on Serialization if this deprecation is removed
--- a/src/check_diagnostics.jl
+++ b/src/check_diagnostics.jl
@@ -1,0 +1,75 @@
+function check_div(chn; per_chain=false)
+    divergent = Array(chn[:numerical_error])
+    
+    n_for_chains = sum(divergent, dims=1)
+    
+    n_divergent = sum(n_for_chains)
+    if n_divergent > 0
+        N = length(divergent)
+        @info "Warning: $(n_divergent) of $(N) iterations ended with a divergence ($(100 * n_divergent / N) %)"
+        if per_chain
+            chain_len, num_chains = size(divergent)
+            for chain_num in 1:num_chains
+                if n_for_chains[chain_num] > 0
+                    @info "Chain $(chain_num): $(n_for_chains[chain_num]) of $(chain_len) iterations ended with a divergence ($(100 * n_for_chains[chain_num] / chain_len) %)"
+                end
+            end
+        end
+        
+        @info "Try running with larger target acceptance rate to remove the divergences."
+        return false
+    else
+        return true
+    end
+end
+
+function check_n_eff(chn)
+    no_warning = true
+    
+    n_steps = size(chn, 1)
+    ess_df = ess(chn)
+    for ix in size(ess_df)[1]
+        param_name = string(ess_df[ix,:parameters])
+        n_eff = ess_df[ix,:ess]
+        ratio = n_eff / n_steps
+        if ratio < 0.1 || isnan(ratio) || isinf(ratio)
+            no_warning = false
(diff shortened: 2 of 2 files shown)


devmotion:
My initial impression is that these functions do not belong into MCMCChains since they assume the existence of certain metadata that is not present in chains in general. In particular, `:numerical_error` is only present when you use Turing's NUTS sampler. Maybe one could include checks of ess and rhat since they are computed by MCMCChains. However, I have the feeling it would be easier for users just to perform these simple checks (e.g., 0.9 < rhat < 1.1) themselves since one might want to adapt it anyways, e.g., check only some samples, parameters, per chain, with a specific ESS algorithm etc., and so either the generic implementation becomes messy or users will just reimplement it anyway.

So in general, I think it would be better to mention in the documentation prominently how users can analyze the quality of their samples.

devmotion:
Maybe a simple alternative would be to add a `verbose` keyword argument to the `ess` function and add warnings or infos if verbose output is requested (by default it would be false). This would also avoid to recompute ess and rhat twice.

treigerm:
As a Turing user I would still like to have a simple function which does some simple sanity checks on all the diagnostics at once and then I can investigate any issues more closely using custom functions and plots. Ideally, Turing would run these checks automatically for me when I sample using NUTS. As far as I know both Stan and PyMC3 do that. That way I know immediately when something is wrong. Also it's useful for beginners (like me) who are just starting out to learn about Bayesian modeling. They're unlikely to already have read through all the different possible diagnostics (for example, I still haven't really understood the E-BFMI diagnostic) but then encountering specific warnings might encourage people more to read through some background material. Also I would think that many beginners might simply use Turing's `sample` function as a black-box that gives posterior samples and they only want to learn the minimum of what they need about the specifics of HMC/NUTS. So some people then might be using bad posterior samples without them knowing if there are no warning.

Sorry this got a bit longer than I thought. My main point being that as a user I would really like these diagnostics to somewhere easily accessible if I am loading Turing and I think it would also be useful for others. Does that make sense? Also if MCMCChains is not the best place for these diagnostics could they be put somewhere else?

devmotion:
I definitely agree that it is interesting for (new) users to know if there are any problems with the samples. AdvancedHMC outputs warnings if there are any numerical issues, so users are already notified about these issues. I am a bit sceptical if there should be warning about specific statistics (some simple analysis is already displayed when chains are printed in the REPL). All these estimators such as ess or rhat are only heuristics (and as you say there exist many more) and depend on many different settings (there are tons of options and alternatives for ess alone that will lead to different estimates). The results will be different if you discard initial samples, thin the samples etc. So these simple numbers can be misleading, there would be false-positive warnings and false-negative non-warnings, and the analysis will be much less sophisticated and fine-tuned to your problem than any custom call of `ess` etc that you can perform afterwards. In general, in my opinion, users should always plot the chains and inspect the output of `sample` and never just use it as part of a black-box analysis.

On a side note, evaluating these statistics can be computationally demanding, see e.g. benchmarks of ess with different algorithms in https://github.com/TuringLang/MCMCChains.jl/pull/156. This is another reason why it might be better to not compute such statistics as part of the Turing sampling pipeline but instead let users perform a fine-tuned and more performant and feasible analysis separately. Also it doesn't seem to fit well with Turing's ability to pause and resume sampling.

treigerm:
Fair enough, that makes sense. Thanks for the feedback!

cpfiffer:
I will say that I agree that we should have some kind of Turing-side diagnostics (perhaps enabled with a keyword) that runs some checks we know are possible. I kind of like the stuff that @treigerm added here and it'd be nice to have available to the Turing community at-large. Maybe a better home for this is in Turing, with `sample(...; check_diagnostics=true)` running this check after chain construction?

cpfiffer:
But I also agree that this should not live in MCMCChains, to be clear. David's right on that one.

devmotion:
Another alternative: maybe one could colorize the display of `Chains` in the REPL and mark low and "bad" rhat values in red (or rather in a special color from the user defined colorscheme). Mainly since these standard evaluations of ESS and rhat are performed anyway and we would not have to add a keyword argument to `sample` (there are so many already...).

cpfiffer:
No, it shouldn't be color-based. We have to assume that users are using plain text displays or `julia` without color for displaying statistical information, which means either a separate diagnostic report, a new column, or something else.

treigerm:
I opened a new issue in Turing just so that this discussion get lost in a closed PR.

PR #263, opened 2021-02-13
Return analytical distribution of R star statistic

devmotion:
I noticed that the distribution of the R star statistic in Algorithm 2 of the original paper can be calculated analytically and one does not have to resort to empirical samples for probabilistic predictions, as proposed in the publication. I found some time yesterday and played around with it a bit. This PR contains an implementation of the idea and adds tests also for a simple deterministic classifier.

Since the change is breaking, the PR bumps the major version and also removes existing deprecations. The implementation uses `Distributions.poissonbinomial_pdf` and `Distributions.Dirac` which require Distributions 0.24.2 and 0.24.7, respectively.


cpfiffer (approved):

yebai:
@devmotion This looks potentially worth more than a PR. I'd encourage you to write a short tech note for this. 

@trappmartin Would you like to take a look at this PR before it gets merged?

trappmartin:
Really nice. Yes I would definitely like to have a look.

src/rstar.jl
```diff
@@ -84,12 +87,22 @@ function rstar(rng::Random.AbstractRNG, classif::MLJModelInterface.Supervised, c
...
 
-function rstar_score(rng::Random.AbstractRNG, classif::MLJModelInterface.Deterministic, fitresult, xtest, ytest)
-    pred = MLJModelInterface.predict(classif, fitresult, xtest)
-    return mean(((p,y),) -> p == y, zip(pred, ytest))
+function rstar_distribution(
+    pred::AbstractVector{<:UnivariateDistribution},
+    ytest::AbstractVector,
+)
+    # Compute probabilities of Poisson Binomial distribution with support `0:length(ytest)`.
+    probs = Distributions.poissonbinomial_pdf(map(pdf, pred, ytest))
```
devmotion, line 103: This algorithm has quadratic complexity in the number of test data points. However, the mean and variance of the distribution can be computed from the predicted probabilities with linear complexity.

devmotion:
This PR motivated two PRs to Distributions, https://github.com/JuliaStats/Distributions.jl/pull/1285 and https://github.com/JuliaStats/Distributions.jl/pull/1286. If these changes are available, we could exploit the mathematical structure of the resulting distribution better and return a scaled PoissonBinomial distribution. This would allow us to skip the computation of the full probability mass function (the complexity is quadratic in the number of trials, i.e., number of test samples here) when e.g. only the mean and variance are of interest (which can be computed in linear time).

Alternatively, one could implement a custom distribution (i.e., basically reimplement PoissonBinomial) but this seems suboptimal.

More generally, I wonder if it would make sense to scale the statistic from the interval `[0, nchains]` to `[0, 1]` - it seems this would simplify the comparison of estimates for chains with different number of chains and possibly increase the interpretability?

devmotion:
Closed in favour of https://github.com/devmotion/InferenceDiagnostics.jl/pull/7.

PR #265, opened 2021-02-22
Add Documenter.jl

rikhuijzer:
As briefly discussed in https://github.com/TuringLang/Turing.jl/issues/1474, this PR gives setting up Documenter.jl for MCMCChains.jl a shot. I had to make quite a lot of decisions to make reading the docs more intuitive. In summary, I've aimed to

- move as much text as possible into the docstrings, because that way it can be included in the docs (via, for example, ```@docs foo ```), users can see it via `help?> foo`, and it's closer to the actual code so less likely to become [outdated](https://blog.royalsloth.eu/posts/on-navigating-a-large-codebase/#documentation-problems).
- mention all exported functions somewhere in the docs; the only exception currently is `ChainDataFrame`.

### DocStringsExtensions

I've tried DocStringsExtensions to avoid some duplication, but it introduced too much noise unfortunately: Instead of 
```
rstar([rng ,]classif::Supervised, chains::Chains; kwargs...)
rstar([rng ,]classif::Supervised, x::AbstractMatrix, y::AbstractVector; kwargs...)
```
`$TYPEDSIGNATURES` will generate
```
rstar(rng::Random.AbstractRNG, classif::MLJModelInterface.Supervised, x::AbstractMatrix{T} where T, y::AbstractVector{Int64}; iterations, subset, verbosity)
```
and for 
```
rafterydiag(x::Vector{<:Real}; q = 0.025, r = 0.005, s = 0.95, eps = 0.001, range = 1:length(x))
```
it will generate
```
rafterydiag(x::Vector{var"#s48"} where var"#s48"<:Real; q, r, s, eps, range) -> Vector{Float64}

### CODE REVIEW

**Comment by rikhuijzer** [comment]
Of course, I forgot the most important information. The preview is available at <https://rikhuijzer.github.io/MCMCChains.jl/dev/>.

File: docs/make.jl [hunk 1]
```diff
@@ -0,0 +1,25 @@
+push!(LOAD_PATH,"../src/")
+
```
devmotion, line 2: This should be removed:
```suggestion
```
Instead, the package should be added in the standard way (as done in the Github action).

docs/make.jl
```diff
@@ -0,0 +1,25 @@
...
+using Documenter
+using MCMCChains
+
+name = "MCMCChains.jl"
+
+makedocs(
+    sitename = name,
```
devmotion, line 9: I think `name` should be removed - it is too generic and we will never change the name anyway:
```suggestion
makedocs(
    sitename = "MCMCChains.jl",
```

docs/make.jl
```diff
@@ -0,0 +1,25 @@
...
+            "Chains" => "chains.md",
+            "Diagnostics" => "diagnostics.md",
+            "Posterior statistics" => "stats.md",
+            "Model selection" => "modelstats.md"
+        ]
+    ],
+    format = Documenter.HTML(prettyurls = get(ENV, "CI", nothing) == "true")
+)
+
+deploydocs(repo = "github.com/rikhuijzer/$name.git", devbranch = "master")
```
devmotion, line 25: The URL is incorrect and `devbranch` is set to the default value. I think it would also be good to build previews: ```suggestion
deploydocs(;
    repo="github.com/TuringLang/MCMCChains.jl.git", push_preview=true
)
```
We should also add an action that cleans the previews (you can find an example in the Documenter docs and https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/blob/master/.github/workflows/doc_preview_cleanup.yml).

File: docs/src/index.md [hunk 4]
```diff
@@ -0,0 +1,232 @@
...
+chn = Chains(...)
+chn_param1 = chn[:,2,:] # returns a new Chains object for parameter 2
+chn[:,2,:] = ... # set values for parameter 2
+```
+
+## Parameter Names
+
+Chains can be constructed with parameter names, like so:
+
+```julia
```
devmotion, line 37: It would be good to make these proper examples.

devmotion: The same is true for all the code here.

src/chains.jl
```diff
@@ -86,6 +86,15 @@ Chains(chain::Chains, ::Nothing) = chain
 
 # Groups of parameters
 
+"""
+    namesingroup(chains::Chains, sym::Union{String,Symbol})
+
+Returns the names of all parameters in a chain that belong to the group `:name`.
```
devmotion, line 92: Docstrings should start with a single one-line sentence that describes the function in the imperative form (see https://docs.julialang.org/en/v1/manual/documentation/):
```suggestion
Return the names of all parameters in a chain that belong to the group `:name`.

```

src/chains.jl
```diff
@@ -86,6 +86,15 @@ Chains(chain::Chains, ::Nothing) = chain
...
+This is based on the MCMCChains convention that parameters with names of the form `name[index]`
```
devmotion, line 93: Maybe
```suggestion
This is based on the MCMCChains convention that parameters with names of the form `:name[index]` 
```
?

src/chains.jl
```diff
@@ -95,6 +104,11 @@ function namesingroup(chains::Chains, sym::Symbol)
     return names_of_params[indices]
 end
 
+"""
+    group(chains::Chains, name::Union{String,Symbol}
```
devmotion, line 108:
```suggestion
    group(chains::Chains, name::Union{String,Symbol})
```

src/chains.jl
```diff
@@ -95,6 +104,11 @@ function namesingroup(chains::Chains, sym::Symbol)
...
+
+Returns a subset of the chain chain with all parameters in the group `:name`.
```
devmotion, line 110: The same applies here (and below):

```suggestion
Return a subset of the chain chain with all parameters in the group `:name`.
```

devmotion, line 110: Though probably

```suggestion
Return a subset of the chain chain with all parameters in the group `Symbol(name)`.
```
is better here.

src/chains.jl
```diff
@@ -612,7 +626,26 @@ function pool_chain(c::Chains)
...
 end
 
-# replace parameter names
+"""
+    replacenames(chains::Chains, dict::AbstractDict) 
+
+Replace parameter names by creating a new `Chains` object that shares the same underlying data.
+
+# Example
+```
```
devmotion, line 635: This should be written as a proper example, and preferably as a doctest (which should be tested, of course).

src/discretediag.jl
```diff
@@ -27,9 +25,9 @@ function hangartner_inner(Y::AbstractMatrix, m::Int)
 end
 
 """
-  weiss(X::AbstractMatrix) -> (statistic, m_tot, pvalue, ca)
+    weiss(X::AbstractMatrix)
 
-The weiss procedure to assess convergence in MCMC output computes X^2/c and evaluates a p-value from the X^2 distribution with (|R| − 1)(s − 1) degrees of freedom.
+The weiss procedure to assess convergence in MCMC output computes ``\\frac{X^2}{c}`` and evaluates a p-value from the ``X^2`` (Chi-square) distribution with ``(|R| − 1)(s − 1)`` degrees of freedom.
```
devmotion, line 30: Slightly better maybe:
```suggestion
Assess the convergence of the MCMC chains with the Weiss procedure.

It computes ``\\frac{X^2}{c}`` and evaluates a p-value from the ``\\chi^2`` distribution with ``(|R| − 1)(s − 1)`` degrees of freedom.
```

However, it would be good to add a reference and explain all variables.

src/rstar.jl
```diff
@@ -4,8 +4,8 @@
 
 Compute the R* convergence diagnostic of MCMC.
 
-This implementation is an adaption of Algorithm 1 & 2, described in [^LambertVehtari]. Note
-that the correctness of the statistic depends on the convergence of the classifier used
+This implementation is an adaption of Algorithm 1 & 2, described by Lambert & Vehtari ([2020](https://arxiv.org/abs/2003.07900)).
```
devmotion, line 7: Why was this changed? The previous version used the Documenter format for references.

devmotion (changes requested):

.github/workflows/CI.yml
```diff
@@ -56,4 +51,22 @@ jobs:
...
+      - uses: julia-actions/setup-julia@v1
+        with:
+          version: '1'
+      - run: |
+          julia --project=docs -e '
+            using Pkg
+            Pkg.develop(PackageSpec(path=pwd()))
+            Pkg.instantiate()'
+      - run: julia --project=docs docs/make.jl
+        env:
```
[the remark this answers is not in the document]
rikhuijzer: Awesome (and huge) list of comments! This is my favorite since it even fixes the problems I had on my local (NixOS) machine. I now have to work on a paper and will work on this PR again later today or tomorrow.

.github/workflows/CI.yml
```diff
@@ -56,4 +51,27 @@ jobs:
...
+  docs:
+    name: Documentation
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install dependencies for GR backend.
+        # See https://github.com/JuliaPlots/Plots.jl/issues/1649#issuecomment-682147970 for details;
+        # also, see https://gr-framework.org/julia.html#installation.
+        run: |
+          sudo apt-get install -y libxt6 libxrender1 libxext6 libgl1-mesa-glx libqt5widgets5
```
devmotion, line 65: These are heavy dependencies and should not be needed. I've never had to install them in CI. The suggested environment variable (which was also suggested by the developer of GR in the linked issue) should be sufficient.

```suggestion
```

In my experience, plotting also still works, even with the error messages.

devmotion (changes requested):

.github/workflows/CI.yml
```diff
@@ -13,15 +13,10 @@ jobs:
     strategy:
       matrix:
         version:
-          - '1.0'
```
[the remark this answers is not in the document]
rikhuijzer: Given that I expected to commit a dozen of times, I found all these build minutes quite wasteful so temporarily pruned the matrix a bit. In the next commit, they will be back.

src/chains.jl
rikhuijzer: In the next commit, I have run a grep over the repo to solve these kinds of issues.

rikhuijzer:
Could someone with the right permissions set up the DOCUMENTER_KEY? Based on the error
```
fatal: unable to access 'https://github.com/TuringLang/MCMCChains.jl.git/': The requested URL returned error: 403
```
I assume that that is the problem for the Documentation CI job.

src/rstar.jl
rikhuijzer: The references were not consistent; Most used a url and some text and this was the only one which used a footnote. The footnote looks looks like this in the text:

image

and like this in at the bottom of the page:

image

A relevant discussion is also at [discourse](https://discourse.julialang.org/t/citations-in-docs/5066/37). Also, Julia shows them just in text at https://docs.julialang.org/en/v1/base/math/#Base.Math.hypot. Anyway, I don't have a strong opinion on this. Let me know what you prefer and I'll change it to that style.

devmotion: On the same page, https://docs.julialang.org/en/v1/base/math/#Base.Math.evalpoly shows an example where base Julia uses the Documenter-style references :slightly_smiling_face:

README.md
```diff
@@ -3,383 +3,17 @@
 ![CI](https://github.com/TuringLang/MCMCChains.jl/workflows/CI/badge.svg?branch=master)
 [![codecov](https://codecov.io/gh/TuringLang/MCMCChains.jl/branch/master/graph/badge.svg?token=TFxRFbKONS)](https://codecov.io/gh/TuringLang/MCMCChains.jl)
 [![Coverage Status](https://coveralls.io/repos/github/TuringLang/MCMCChains.jl/badge.svg?branch=master)](https://coveralls.io/github/TuringLang/MCMCChains.jl?branch=master)
+[![][docs-stable-img]][docs-stable-url] 
+[![][docs-latest-img]][docs-latest-url]
 
 Implementation of Julia types for summarizing MCMC simulations and utility functions for diagnostics and visualizations.
 
-## Example
```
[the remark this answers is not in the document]
rikhuijzer: I have started with it but got scared of a bit by the introduction of yet another opportunity for outdated docs. Arguably, DataFrames has gotten quite popular even though it [doesn't show examples in its README](https://github.com/JuliaData/DataFrames.jl/blob/main/README.md). If you don't mind, I just add a basic `plot(chn)` and link it to the docs. This will be included in the next commit.

rikhuijzer:
Small summary of all these commits: The doctests were flaky (random pass/fail) during local development and I still haven't been able to pinpoint it exactly. Currently, the doctests don't fail in the documenter run, but they **do** fail in the test runs. Now, the code for the doctests is the same as https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/239.

The next commits will be working on adding more doctests.

rikhuijzer:
I've now applied all the feedback suggestions. One thing I could get to work though was the following code block:

 ```julia
using KernelDensity

# Construct a kernel density estimator
c = kde(Array(chn[:s]))

# Generate 10000 weighted samples from the grid points
chn_weighted_sample = sample(c.x, Weights(c.density), 100000)
```

File: Project.toml [hunk 16]
```diff
@@ -43,3 +43,10 @@ StatsBase = "0.32, 0.33"
 TableTraits = "0.4, 1"
 Tables = "1"
 julia = "1"
+
+# Required for Julia version < 1.2. 
+[extras]
+Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
+
+[targets]
+test = ["Documenter"]
```
devmotion, line 52: We explicitly activate `test/Project.toml` on Julia < 1.2 - is it not sufficient to add Documenter to test/Project.toml?

docs/Project.toml
```diff
@@ -0,0 +1,9 @@
+[deps]
+Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
+MCMCChains = "c7f686f2-ff18-58e9-bc7b-31028e88f75d"
```
devmotion, line 3: It seems the MCMCChains dependency reappeared:
```suggestion
```

Project.toml
rikhuijzer: That's strange. I did that based on the failing action for commit `fdedc01` and the [test/Project.toml source code](https://github.com/rikhuijzer/MCMCChains.jl/blob/fdedc019a9a446ee9a00efd40cbda9c42120f353/test/Project.toml).

devmotion:
> One thing I couldn't get to work though was the following code block:

It seems the following should work:
https://github.com/TuringLang/MCMCChains.jl/blob/9dcd97999404e9694ebef905acf47616a84775a7/test/sampling_tests.jl#L20-L22

I noticed that I have not sufficient permissions to add the DOCUMENTER_KEY. @cpfiffer, could you add one?

docs/Project.toml
rikhuijzer: My bad. Fixed again.

Project.toml
rikhuijzer: Related: I've changed continue-on-error to
```
continue-on-error: ${{ matrix.version == 'nightly' || matrix.os == 'ubuntu-latest' }}
```
because the CI sometimes fails on a test in rstar.jl even though I didn't touch the source code. I hope that allowing the Ubuntu tests to continue will show what's up with that.

rikhuijzer:
> It seems the following should work:
> 
> https://github.com/TuringLang/MCMCChains.jl/blob/9dcd97999404e9694ebef905acf47616a84775a7/test/sampling_tests.jl#L20-L22

Thanks. I noticed that this required quite a lot of lines of code to show in the docs because of all the dependencies. Therefore, I've referred users to the test now.

cpfiffer:
> I noticed that I have not sufficient permissions to add the DOCUMENTER_KEY. @cpfiffer, could you add one?

Added, I think. I also made you an admin since it's kind of dumb that you don't have perms here.

rikhuijzer:
Thanks, @cpfiffer. To avoid people having to go into the logs as well: the error is still the same. Are the write permissions set for the key?

devmotion:
The permissions seem to be correct.

.github/workflows/CI.yml
```diff
@@ -9,7 +9,7 @@ on:
 jobs:
   test:
     runs-on: ${{ matrix.os }}
-    continue-on-error: ${{ matrix.version == 'nightly' }}
+    continue-on-error: ${{ matrix.version == 'nightly' || matrix.os == 'ubuntu-latest' }}
```
devmotion, line 12: This should be reverted as well before the PR is merged (I assumed it was changed for debugging):
```suggestion
    continue-on-error: ${{ matrix.version == 'nightly' }}
```

devmotion:
It seems the plotting tests fail on Julia 1.0 due to a StackOverflowError - I wonder if it is caused by some upstream changes since this PR doesn't touch them, it seems.

.github/workflows/CI.yml
rikhuijzer: Done :+1:

devmotion:
The docs can't be deployed since your PR is made from a fork (https://juliadocs.github.io/Documenter.jl/stable/lib/public/#Documenter.deploydocs):

> Pull requests made from forks will not have previews. Hosting previews requires access to the deploy key. Therefore, previews are available only for pull requests that were submitted directly from the main repository.

devmotion:
The StackOverflowError seems to be caused by OffsetArrays: https://github.com/JuliaArrays/OffsetArrays.jl/issues/209

Project.toml
rikhuijzer: The discussion at <https://github.com/JuliaLang/Pkg.jl/issues/2097#issuecomment-714316161> seems related. How about keeping the workaround with `[extras]` for a bit? It's likely that 1.6 is gonna be LTS anyway according to https://discourse.julialang.org/t/julia-v1-6-0-beta1-is-now-available/53065/5.

rikhuijzer:
> The docs can't be deployed since your PR is made from a fork (https://juliadocs.github.io/Documenter.jl/stable/lib/public/#Documenter.deploydocs):
> 
> > Pull requests made from forks will not have previews. Hosting previews requires access to the deploy key. Therefore, previews are available only for pull requests that were submitted directly from the main repository.

Obvious in hindsight. I naively hoped that GitHub had a workaround for this. Anyway, thanks for looking it up.

> The StackOverflowError seems to be caused by OffsetArrays: JuliaArrays/OffsetArrays.jl#209

Good catch. Someone is already working on a PR there.

Is there anything more I can fix?

Project.toml
devmotion: I fixed the test setup, the only issue was that you tried to load Documenter in runtests.jl before test/Project.toml was activated.

devmotion: It is not relevant if 1.6 becomes the new LTS (which will also only happen - if it happens - when 1.7 is released), we should always test the oldest version that is supported. I.e. it can only be changed if we drop support for Julia < 1.2.

src/rstar.jl
```diff
@@ -21,17 +23,16 @@ verbosity level.
...
-# Compute 20 samples of the R* statistic using sampling from according to the prediction probabilities.
+# Compute 20 samples of the R* statistic using sampling according to the prediction probabilities.
 Rs = rstar(XGBoost(), chn; iterations=20)
 
 # estimate Rstar
 R = mean(Rs)
 
 # visualize distribution
+using StatsPlots
 histogram(Rs)
```
devmotion, line 34: I don't think it is helpful to use `StatsPlots` or `histogram` here in the documentation. In principle, on Julia >= 1.3, this could be made a doctest otherwise.

devmotion (commented):
I added some additional, mostly minor comments. Apart from these we have to wait for the fix in OffsetArrays. I am confident that it will be available soon since a PR already exists.

docs/src/statsplots.md
```diff
@@ -0,0 +1,102 @@
...
+autocorplot(chn)
+```
+
+## Corner
+
+For some reason, this one fails
+
+```@example statsplots
+# corner(chn)
+```
```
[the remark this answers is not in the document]
rikhuijzer: https://github.com/TuringLang/MCMCChains.jl/issues/269

docs/src/index.md
```diff
@@ -0,0 +1,179 @@
...
+
+Note that only `:a` and `:b` are being shown. You can explicity retrieve
+an array of the summary statistics and the quantiles of the `:internals` section by
+calling `describe(chn; sections = :internals)`, or of all variables with
+`describe(chn; sections = nothing)`. Many functions such as [`MCMCChains.summarize`](@ref) or
+[`MCMCChains.gelmandiag`](@ref) support the `sections` keyword argument.
+
+## Groups of parameters
+
+You can access the names of all parameters in a `chain` that belong to the group `:name` by using
```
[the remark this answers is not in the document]
rikhuijzer: I've tried to implement the second point, but wasn't able to use `name` instead of `:name` everywhere. At some points, I was. But at other points, it lead to two problems:
1. What should be the content of variable `a` be? For example, with `a = :A`, it is still slightly more confusing for the reader when comparing the code to the output.
2. The code is harder to copy paste, because the snippets are less stand alone.

For now, I'm changing it to where possible.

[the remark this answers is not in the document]
rikhuijzer: > The signature of the docstring contains only sym - we should make this consistent and probably use something like group everywhere here (also in the docstring and the signature).

I've also tried this, but found a problem here too. There is a function called `group` below `namesingroup`. If I change the parameter in `namesingroup` to `group`, then I should change the parameter of `group` to `group` too for consistency.

docs/src/statsplots.md
```diff
@@ -0,0 +1,102 @@
...
+
+\
+
+```@example statsplots
+p2 = plot(chn, colordim = :parameter)
+# Using png to avoid vectorized graphics for images with many vectors. # hide
+plot!(size=(840, 400), fmt = :png) # hide
+```
+
+\
```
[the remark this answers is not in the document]
rikhuijzer: > e.g. shorten the chains

So simple. Great solution :+1:

src/chains.jl
```diff
@@ -86,6 +107,26 @@ Chains(chain::Chains, ::Nothing) = chain
...
 
+"""
+    namesingroup(chains::Chains, sym::Union{String,Symbol})
+
+Return the names of all parameters in a chain that belong to the group `:name`.
+
+This is based on the MCMCChains convention that parameters with names of the form `:name[index]` 
+belong to one group of parameters called `:name`.
+
+If the chain contains a parameter of name `:name` it will be returned as well.
```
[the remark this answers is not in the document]
rikhuijzer: For clarity, I listed some problems with this suggestion above.

rikhuijzer:
Thank you, David, for the extensive feedback again! Nice rule-of-thumb about the return type in docstrings. My decision was indeed a bit arbitrary. Also, thanks for fixing the code loading in `test/runtests.jl`; now I see what you meant :eyes:.

Summary of the changes:
- Applied most feedback; let me know if you don't agree with the "problems" that I have raised.
- jldoctests now fail when running make.jl (thanks to Morten Piibeleht https://github.com/JuliaDocs/Documenter.jl/issues/1535).
- The OffsetArrays patch is merged, but no new version was tagged. I've asked them to register the patch with the registries.

We still have to do something about `dic`, or consider it out of the scope for this PR. The broken example is still in the docstring (https://github.com/TuringLang/MCMCChains.jl/issues/267).

rikhuijzer:
Arg. Those doctests are difficult to get right.

rikhuijzer:
The rstar doctest failed on a warning which was printed to the output. Since jldoctest tests the raw string output, this failed. As far as I know, Julia has no simpler way to catch stdout/stderr than something like
```
io = IOBuffer()
redirect_stderr(outf) do
...
end
```
Also, I don't see a way to do a smoke test in Documenter, that is, to run the code but ignore the output. So, I just removed the jldoctest and referred readers to the test.

devmotion:
I think `dic` should be fixed in a separate PR.

> Also, I don't see a way to do a smoke test in Documenter, that is, to run the code but ignore the output. So, I just removed the jldoctest and referred readers to the test.

You can indicate parts of the output that should be ignored with `[...]`. From the [Julia documentation](https://docs.julialang.org/en/v1/manual/documentation/):

> To indicate that the output result is truncated, you may write `[...]` at the line where checking should stop. This is useful to hide a stacktrace (which contains non-permanent references to lines of julia code) when the doctest shows that an exception is thrown, for example

devmotion:
BTW `corner` works for me - maybe you used an old version of StatsPlots?

rikhuijzer:
> You can indicate parts of the output that should be ignored with [...]. From the Julia documentation:

I had to use another approach but this is now solved :+1:. 

> BTW corner works for me - maybe you used an old version of StatsPlots?

Solved by not using `theme(:ggplot2)`.

rikhuijzer:
@devmotion is there anything that I can do to move this PR forward?

devmotion:
I'll have another look later today :+1:

devmotion (approved):
Looks good to me!

Thanks a lot @rikhuijzer for this PR! It's great to finally have some more proper documentation of MCMCChains and doctests!

rikhuijzer:
@devmotion Thanks for the merge and all the feedback. It was a bit of struggle, but I've learned a lot about writing software and MCMCChains. :+1: Totally worth it

PR #271, opened 2021-03-03
Add DIC

itsdfish:
Hi all,

I worked with Rob to create a library of model comparison metrics at [StatsModelComparisons](https://github.com/StatisticalRethinkingJulia/StatsModelComparisons.jl). This package has a generic interface for using model comparison metrics. This PR provides a method for computing DIC from a chain object. Please let me know if there are any edits you would like me to make.


itsdfish:
It looks like this failed due to a compatibility issue. MCMCChains tests on Julia 1.0.5, but [StanDump](https://github.com/tpapp/StanDump.jl/blob/5deab0c727a2d47934e03c5bc34cac36d8afa40c/Project.toml#L6) requires Julia 1.1 or higher. What is the best path forward?

devmotion:
IMO it is a bit problematic that the method requires a parameter `:lp` with the log-likelihood. The name is not standardized, i.e., it might not exist or it might denote something different such as the log joint probability or something else, depending on the sampling algorithm. Probably this was also the reason for the current implementation which requires a log-likelihood function.

Maybe it would be better to remove `dic` from MCMCChains and refer users to other packages such as StatsModelComparisons. Alternatively, one could fix the current implementation and add a simple example in the docstring, e.g., for the case where `:lp` is actually the log-likelihood.

itsdfish:
Part of the rationale for creating `StatsModelComparisons` was to make  model comparison and MCMC output more modular. One approach would be to add methods to each participating package to preprocess and interface with the primary method defined in `StatsModelComparisons`. An alternative approach would be to define some of those connecting methods in `StatsModelComparisons`. The third option is to define your own methods for DIC, WAIC etc. in `MCMCChains` to do all the calculations. I think option 1 or 2 is preferable because it makes the ecosystem more standardized, modular and reduces redundancy, but of course you should choose the option that you think fits best with the goals of `MCMCChains`.

Thanks for pointing out the issue with `lp`.  I also saw `log_density` in the chain output. It's not clear to me why there are two variables that give the same values in most cases. Would it be better to change `lp` to `log_density`? I am going to ping @goedman  so he can participate in these discussions.

devmotion:
I totally agree that it would be good to make the statistical methods more modular - this was also the main motivation for changing (most) implementations in MCMCChains to operate on arrays instead of `Chain` objects recently.

> Thanks for pointing out the issue with lp. I also saw log_density in the chain output. It's not clear to me why there are two variables that give the same values in most cases. Would it be better to change lp to log_density? I am going to ping @goedman so he can participate in these discussions.

Using a different name does not resolve the general issue here. MCMCChains does not enforce any specific parameter name for log-likelihood values (if they exist at all) and lets you rename all parameters arbitrarily. So anything that is based on the evaluation of a specific parameter name is likely to be broken - it will error if it does not exist and, even worse, it will yield completely incorrect results if a parameter of this name exists but does not denote log-likelihood values.

itsdfish:
Ok. I understand the issue now. I was suffering from a bit of myopia. I usually use MCMCChains with specific samplers in Turing. SO the posterior log likelihoods have a specific name. In that case, the simplest solution is probably adding an additional argument for the variable name:

`function dic(chain::Chains, post_log_like::Symbol)
    lps = get(chain, post_log_like).post_log_like[:] |> Array
    return dic(lps)
end`

Do you think something like that would work?

devmotion:
Yep, I think that sounds good. I would just suggest
```julia
function dic(chain::Chains, loglik::Symbol)
    lps = Array(chain[:, loglik, :])
    return dic(lps)
end
```
instead.

itsdfish:
Thanks. I like that better. I modified your method so that `lp` is a vector rather than a matrix.

src/modelstats.jl
```diff
@@ -3,47 +3,18 @@ export dic
...
-    ED = mean(D)
-
-    pD = 2*(ED - EθD)
-
-    DIC = EθD + pD
-
-    return DIC, pD
+function dic(chain::Chains, loglik::Symbol)
+    lps = Array(chain[:, loglik, :][:])
+    return dic(lps)
```
devmotion, line 19: If you need a vector, it's more efficient to write
```suggestion
    lps = Array(chain[:, loglik, :])
    return dic(vec(lps))
```
This does not allocate another array.

devmotion:
> MCMCChains tests on Julia 1.0.5, but StanDump requires Julia 1.1 or higher. What is the best path forward?

Do you know why StanDump requires Julia 1.1?

src/modelstats.jl
itsdfish: Thanks! I pushed the changes

itsdfish:
I'm not sure. Let me find out.

tpapp:
As noted in tpapp/StanDump.jl#7, I don't quite remember, but if you **really** need 1.0, PRs are welcome.

That said, do you really need 1.0?

devmotion:
I personally don't need 1.0 (or anything apart from the latest version) but I think it would be unfortunate to drop Julia 1.0 if it can be supported without much effort. Currently, it does not lead to any additional maintenance burden. AFAIK Julia 1.6 might become the new LTS version only after Julia 1.7 is released, so I'm not sure how long it will take until a new LTS version is available.

tpapp:
> it does not lead to any additional maintenance burden

Let's see — again, PRs are welcome.

Generally I think that so much happened since 1.0 in terms of new features that it does not make sense to support it.

goedman:
The current version of StatsModelComparisons.jl uses StanSample.jl for testing purposes. Wouldn't that be an ongoing source of issues if StatsModelComparisons.jl is made part of testing MCMCChains.jl.

In the past there was a link between Turing and CmdStan for comparisons. I'm no longer sure that's still the case, but if so, that could also be problematic as it would mix and match CmdStan and StanSample.

If there are good reasons to go this route I would suggest to redo the StatsModelComparisons.jl tests using Turing.jl or DynamicHMC.jl.

In the PR, in src/MCMCChains.jl line 27: `import ModelComparisons: dic` uses the initial proposed package name.

itsdfish:
"In the PR, in src/MCMCChains.jl line 27: import ModelComparisons: dic uses the initial proposed package name."

Good catch. I fixed the import statement. 

"The current version of StatsModelComparisons.jl uses StanSample.jl for testing purposes."

Would this issue be resolved if you specified StanSample in extras instead of compat? It was not clear to me based on the documentation for Pkg.jl.

devmotion:
> The current version of StatsModelComparisons.jl uses StanSample.jl for testing purposes.

I.e., it is used only in the tests? Then it should not be listed among the dependencies but in an `[extras]` section (https://julialang.github.io/Pkg.jl/v1/creating-packages/#Test-specific-dependencies-in-Julia-1.0-and-1.1).

itsdfish:
@devmotion, I noticed that too. Would the conflict be resolved if listed under extras?

devmotion:
Yes, since the packages listed under `extras` are not pulled in and installed when the package is installed - they are only used for testing.

goedman:
Just to make sure I understand this right. For the import statement in MCMCChains.jl, StatsModelsComparisons.jl needs to be in the dependencies of MCMCChains right?

In StatsModelsComparisons.jl I can turn all examples into `tests` and then make StanSample.jl a Test dependency.

Interesting!

devmotion:
I opened a PR that also uses RecipeBase instead of StatsPlots: https://github.com/StatisticalRethinkingJulia/StatsModelComparisons.jl/pull/4

itsdfish:
Rob, I believe that is the case.

goedman:
Very nice changes David! Thanks a lot. I'll remove the mc_path. At some point I ran in troubles with making it a constant but that was a few years ago. For this package it is not needed.

Alvaro definitely used the R and Python sources to do the translations after permission by the author (Aki Vehtari I think used Matlab). I will indeed update the LICENSE.md file.

Project.toml
```diff
@@ -40,6 +41,7 @@ PrettyTables = "0.9, 0.10, 0.11"
 RecipesBase = "0.7, 0.8, 1.0"
 SpecialFunctions = "^0.8, 0.9, 0.10, 1.0"
 StatsBase = "0.32, 0.33"
+StatsModelComparisons = "0.1.0"
```
devmotion, line 44: To avoid pulling in too many dependencies:
```suggestion
StatsModelComparisons = "0.1.1"
```

src/MCMCChains.jl
```diff
@@ -24,6 +24,8 @@ import Random
 import Serialization
 import Statistics: std, cor, mean, var, mean!
 
+import StatsModelComparisons: dic
+
```
devmotion, line 28: I'd prefer if `dic` is not imported but qualified when implemented below:
```suggestion
import StatsModelComparisons

```

src/modelstats.jl
```diff
@@ -3,47 +3,18 @@ export dic
 #################### Posterior Statistics ####################
 
 """
-    dic(chain::Chains, logpdf::Function) -> (DIC, pD)
+    dic(chain::Chains)
 
-Compute the deviance information criterion (DIC).
-(Smaller is better)
+Computes Deviance Information Criterion (DIC).
```
devmotion, line 8: The first one-line description of the method [should use the imperative form](https://docs.julialang.org/en/v1/manual/documentation/):
```suggestion
Compute the deviance information criterion (DIC).
```

src/modelstats.jl
```diff
@@ -3,47 +3,18 @@ export dic
...
-    lp = zeros(niter + nchains) # resulting logpdf values
-    for i = 1:nparams
-        lp += map(p -> logpdf( ... , x), Array(chain[:,i,:]))
-    end
-    return lp
-end
-
-DIC, pD = dic(chn, lpfun)
-```
+# Arguments
```
devmotion, line 10: This seems to be useful information, maybe it should be added to StatsModelComparisons (as well)?
```suggestion
Note: DIC assumes that the posterior distribution is approx. multivariate Gaussian and tends to select overfitted models.

# Arguments
```

src/modelstats.jl
```diff
@@ -3,47 +3,18 @@ export dic
 #################### Posterior Statistics ####################
 
 """
-    dic(chain::Chains, logpdf::Function) -> (DIC, pD)
+    dic(chain::Chains)
```
devmotion, line 6:
```suggestion
    dic(chain::Chains, loglik::Symbol)
```

src/modelstats.jl
```diff
@@ -3,47 +3,18 @@ export dic
...
-end
-
-DIC, pD = dic(chn, lpfun)
-```
+# Arguments
+* `chain::Chains: an MCMC `Chain` object containing posterior log likelihood values
+* `loglik::Symbol`: variable name associated with posterior log likelihood values
 
+# Returns
+* `dic::Real`: DIC value
```
devmotion, line 15: This seems redundant and is already explained by the first sentence:
```suggestion
```

src/modelstats.jl
```diff
@@ -3,47 +3,18 @@ export dic
...
-        lp += map(p -> logpdf( ... , x), Array(chain[:,i,:]))
-    end
-    return lp
-end
-
-DIC, pD = dic(chn, lpfun)
-```
+# Arguments
+* `chain::Chains: an MCMC `Chain` object containing posterior log likelihood values
+* `loglik::Symbol`: variable name associated with posterior log likelihood values
```
devmotion, line 12: The convention in MCMCChains (and [in the Julia guidelines](https://docs.julialang.org/en/v1/manual/documentation/)) is to not provide a list of arguments but to integrate them in the description if possible. Could you change this?

src/modelstats.jl
```diff
@@ -3,47 +3,18 @@ export dic
...
-
-    D = -2*logpdf(chain)
-    ED = mean(D)
-
-    pD = 2*(ED - EθD)
-
-    DIC = EθD + pD
-
-    return DIC, pD
+function dic(chain::Chains, loglik::Symbol)
```
devmotion, line 17: As mentioned above, I'd prefer
```suggestion
function StatsModelComparisons.dic(chain::Chains, loglik::Symbol)
```

devmotion (changes requested):
I added some suggestions.

src/modelstats.jl
itsdfish: I updated the docs. Sorry. I was not aware of the Julia guidelines.

src/modelstats.jl
```diff
@@ -3,47 +3,13 @@ export dic
 #################### Posterior Statistics ####################
 
 """
-    dic(chain::Chains, logpdf::Function) -> (DIC, pD)
+    dic(chain::Chains, loglik::Symbol)
 
-Compute the deviance information criterion (DIC).
-(Smaller is better)
+Compute the deviance information criterion (DIC) from Chain object on posterior log likelihood samples specified by varible name loglik.
```
devmotion, line 8:
```suggestion
Compute the deviance information criterion (DIC) from `chain` on posterior log likelihood samples specified by parameter name `loglik`.
```

devmotion (commented):
Looks good to me :+1: I am not sure what's the best way to test the implementation but maybe we could at least call it with some random data to make sure it is not completely broken?

itsdfish:
Yeah. That is a good idea. I also added a test to ensure that it fails as expected too. `dic` was no longer visible after your changes. I added a using and export statement. I'm not sure how you want to handle that. importing dic should prevent other variables in StatsModelComparisons from being exported in the MCMCChains.

devmotion:
I noticed that in the mean time the license of StatsModelComparisons was changed to GPL-3 - does this mean we would have to relicense `MCMCChains.dic` (or maybe even other parts?) as well? This seems quite unfortunate, in particular since `StatsModelComparisons.dic` is a completely straightforward Julia implementation and only `StatsModelComparisons.psisloo` was derived from a GPL-3 licensed MATLAB implementation.

itsdfish:
Based on our previous conversation, I was under the impression that the GPL license would only apply to the code for psisloo methods. @goedman, did your plans change?

goedman:
Yes, this recent change to only auto-merge packages with OSI-approved licenses was a surprise to me as well. A problem is that TuringLang is a product while StatisticalRethinkingJulia is just for learning purposes. I have derived the WAIC part from the SR implementation so legally it is probably also GPL-ish. But I'm definitely open to change the license file if that is possible and would work for TuringLang.

devmotion:
@cpfiffer What's your opinion about the license issue?

cpfiffer:
Dunno, I'm kind of the wrong person to ask -- I don't know much about licensing issues.

devmotion:
Unfortunately, I also don't know much about these problems. I would like to fix `dic` and merge this PR, and I think it is great if there's a general interface for methods such as `dic` that we can implement instead of rolling our own implementations. However, my gut feeling is that the `MCMCChains.dic` would have to be licensed under GPLv3 as well since StatsModelComparisons uses this license (as far as I understand, we don't have to relicense the other parts of MCMCChains). Even though it's just a single function I think it would be unfortunate - the implementation of `dic` is quite trivial and, as far as I understand, in principle it could also use the MIT license in StatsModelComparisons.

I also just noticed that StatsBase already defines `aic`, `bic`, and `aicc` (it seems they are not overloaded or reexported in StatsModelComparisons even though the package depends on StatsBase but reimplemented - @itsdfish is this intentional?) but it seems `dic` is missing. Maybe one should add `dic` to StatsBase? Then we could skip the dependency on StatsModelComparisons completely for now and avoid all licensing issues.

itsdfish:
@devmotion, I like your idea of submitting a PR to StatsBase for DIC. I did not know that they implemented other model comparison metrics. 

I will look at the algorithm for PSISLOO to see if it would be feasible to implement. It looks much more involved, so it is probably a long shot for me.

PR #272, opened 2021-03-03
Fix README.md and add a few more doctests

rikhuijzer:
Now that the docs are visible online, I spotted a few minor mistakes and that the image in the README is broken. Also, I've added a few more doctests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ using StatsPlots
 
 plot(chn)
 ```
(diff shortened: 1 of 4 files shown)


rikhuijzer:
Either the doctests are extremely hard to get right, or I'm doing something wrong. Either way, I've removed them for now. At least, it is better than it used to be.

cpfiffer (approved):
Thanks!

PR #275, opened 2021-03-11
Add Gadfly examples

rikhuijzer:
I have been working on moving the Gadfly plotting functionality from my [package](https://github.com/rikhuijzer/TuringPlots.jl) into MCMCChains.jl. However, I figured that I was basically

1. Catching calls to `Gadfly.plot(chn::Chains, args...; vargs..., filter=nothing)`
1. Converting `chn` to `df::DataFrame`
1. Applying `filter` to `df`
1. Invoking `Gadfly.plot(df, args...; vargs...)`

So, most of this could be replaced by using the DataFrames constructor provided by MCMCChains.jl. Since this is easy in use, but can be hard to figure out, I've added a few examples to the docs. 

Some examples from the generated documentation:

image

image

I wonder if parts of these examples should be implemented in `src`. I don't think so given the relative easy of use, but am not sure about it. 

What would possibly be interesting is to implement a function which uses `Gadfly.hstack` to show the density and trace side by side. From my [package](https://rikhuijzer.github.io/TuringPlots.jl/dev/):

image

For the examples, creating such a hstack was a bit too involved.

# Vertical bars for credibility intervals

On a side-note, my package was also showing vertical bars for credibility intervals, like

image

This functionality is hopefully going to be merged into Gadfly, see https://github.com/GiovineItalia/Gadfly.jl/pull/1521.

### DIFF
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -1,13 +1,17 @@
 [deps]
+CategoricalArrays = "324d7699-5711-5eae-9e2f-1d82baa6b597"
 DataFrames = "a93c6f00-e57d-5684-b7b6-d8193f3e46c0"
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
+Gadfly = "c91e804a-d5a3-530f-b6f0-dfbca275c004"
 MLJModels = "d491faf4-2d78-11e9-2867-c94bc002c0b7"
 MLJXGBoostInterface = "54119dfa-1dab-4055-a167-80440f4f7a91"
 StatsPlots = "f3b207a7-027a-5e70-b257-86293d7955fd"
 
 [compat]
+CategoricalArrays = "0.8"
 DataFrames = "0.22"
 Documenter = "0.26"
+Gadfly = "1.3"
 MLJModels = "0.14"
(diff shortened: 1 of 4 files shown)


cpfiffer (approved):
I'm down with it, thanks!

storopoli:
Great addition! Congratulations!

mkborregaard:
are you replacing the StatsPlots functionality with Gadfly here?

rikhuijzer:
> are you replacing the StatsPlots functionality with Gadfly here?

I don't know about the plans of the Turing developers, but I can say that that has never been my intention! I personally like Gadfly, but I also respect people who like Plots.jl. I think both plotting libraries have their strengths and weaknesses.

mkborregaard:
So loading Gadfly here would not break the session for Plots users (due to conflicting definitions of `plot`)?

cpfiffer:
> are you replacing the StatsPlots functionality with Gadfly here?

No, we have no intention of replacing `StatsPlots` with `Gadfly`. This stuff is only loaded if someone really wants to use Gadfly -- we don't enforce any of this in the core package code and probably won't.

PR #279, opened 2021-03-15
Fix implementation of Tables row interface

### LINKED ISSUE #268: Implementation of Tables interface is incorrect
The implementation of the Tables interface is not correct. According to the [documentation](https://tables.juliadata.org/dev/#Implementing-the-Interface-(i.e.-becoming-a-Tables.jl-source)-1), `Tables.rows` has to return an `Tables.AbstractRow`-compatible **iterator**. However, currently it just returns the `Chains` object itself. From the documentation it is unclear to me if the `Tables.AbstractColumns`-compatible object that should be returned by `Tables.columns` should implement the iteration interface as well (it is provided for subtypes of `Tables.AbstractColumns` automatically). 

For instance, Pluto relies on the convention that `Tables.rows` returns an iterator in the implementation of its display mechanism in https://github.com/fonsp/Pluto.jl/blob/6800da3b40a5e38aed465067b38751f6e6479649/src/runner/PlutoRunner.jl#L838. Therefore, currently an error is thrown if one tries to display a `Chains` object in Pluto (e.g., if it is returned from a function call).

sethaxen:
This PR fixes #268 and https://github.com/fonsp/Pluto.jl/issues/868 by correctly implementing `Tables.rows` in terms of `Tables.columntable`.

Because the optional interface methods `Tables.namedtupleiterator` and `Tables.rowtable` were previously implemented to forward to `Tables.columntable` instead of going through the default `Tables.rows` approach, with this change, they no longer are necessary to get the performance boost and have been removed.

Edit: I have verified that Pluto correctly displays `Chains` and `ChainDataFrame` objects with this PR.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -3,7 +3,7 @@ uuid = "c7f686f2-ff18-58e9-bc7b-31028e88f75d"
 keywords = ["markov chain monte carlo", "probablistic programming"]
 license = "MIT"
 desc = "Chain types and utility functions for MCMC simulations."
-version = "4.7.1"
+version = "4.7.2"
 
 [deps]
 AbstractFFTs = "621f4979-c628-5d54-868e-fcf4e3e8185c"
--- a/src/tables.jl
+++ b/src/tables.jl
@@ -43,12 +43,6 @@ Tables.rowaccess(::Type{<:Chains}) = true
 
 Tables.rows(chn::Chains) = Tables.rows(Tables.columntable(chn))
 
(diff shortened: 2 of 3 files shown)


devmotion:
Do the additional tests fail on master and ensure that the implementation is correct now? If not, it would be good to add such tests (maybe one could check that `Table.rows` actually returns an iterator since it seems this was a problem?).

sethaxen:
Good point. The tests were just integration tests. I reorganized the tests and added comprehensive row interface tests that fail on master.

cpfiffer (approved):
I love it, thank you!

PR #284, opened 2021-04-08
Fix failing tests

devmotion:


devmotion:
It might not be possible to set the `displaysize` reliably due to https://github.com/JuliaLang/julia/issues/34721...

src/stats.jl
```diff
@@ -112,7 +112,7 @@ end
 
 function chaindataframe_changerate(name, names_of_params, chains; kwargs...)
     # Compute the change rates.
-    changerates, mvchangerate = chains
+    changerates, mvchangerate = changerate(chains)
```
torfjelde, line 115: Was this just missing by mistake or was there a reason why `changerate` wasn't there before?

src/chains.jl
```diff
@@ -192,14 +192,15 @@ names as the values.
...
+julia> chn = Chains([1:2 3:4]);
 
 julia> get(chn, :param_1)
-(param_1 = [1; 2; … ; 4; 5],)
+(param_1 = [1; 2],)
 
-julia> get(chn, [:param_1, :param_2])
-(param_2 = [6; 7; … ; 9; 10], param_1 = [1; 2; … ; 4; 5])
+julia> get(chn, [:param_2])
+(param_2 = [3; 4],)
```
torfjelde, line 203: I guess this part is how you're dealing with the displaysize-issue you mentioned in the comment?

torfjelde (approved):
I'm no MCMCChains aficionado, but this seems to be a minor change to one particular doctest to deal with the displaysize issue, so I *think* my "LGTM!" sufficies here?

So.

LGTM!

src/stats.jl
devmotion: No it was a bug but we only noticed it since Julia 1.6 performs some more checks when building `NamedTuple`s. Due to this bug fix I incremented the version number.

src/chains.jl
devmotion: Yep, I just reduced the output such that it should work even with tiny displays.

PR #285, opened 2021-04-11
Add ESS/second

cpfiffer:
Updates MCMCChains to allow a user to pass in start and stop times for sampling, to calculate an approximation of ESS per second.

Usage:

```julia
using Revise
using MCMCChains

vs = randn(500, 5, 10)

ts1 = [time() for i in 1:5]
sleep(5)
ts2 = [time() for i in 1:5]

chain = Chains(vs, info = (start_time = ts1,stop_time=ts2))
```

The start, stop, and wall duration estimates are included in the header when `display` is called:

```julia
Chains MCMC chain (500×5×10 Array{Float64, 3}):

Start time        = 2021-04-11T21:04:18.109
Stop time         = 2021-04-11T21:04:23.150
Wall duration     = 5.04 seconds
Iterations        = 1:500
Thinning interval = 1
Chains            = 1, 2, 3, 4, 5, 6, 7, 8, 9, 10
Samples per chain = 500
parameters        = param_1, param_2, param_3, param_4, param_5

Summary Statistics
  parameters      mean       std   naive_se      mcse         ess      rhat   ess_per_sec
      Symbol   Float64   Float64    Float64   Float64     Float64   Float64       Float64

     param_1    0.0276    1.0055     0.0142    0.0156   4942.3626    1.0003      980.4330
     param_2   -0.0038    1.0028     0.0142    0.0146   5093.9868    1.0009     1010.5112
     param_3    0.0147    1.0049     0.0142    0.0127   5043.7801    0.9995     1000.5515
     param_4    0.0190    0.9906     0.0140    0.0143   5103.3872    1.0001     1012.3760
     param_5   -0.0186    0.9796     0.0139    0.0143   5117.6994    0.9993     1015.2151

Quantiles
  parameters      2.5%     25.0%     50.0%     75.0%     97.5%
      Symbol   Float64   Float64   Float64   Float64   Float64

     param_1   -1.9177   -0.6488    0.0244    0.6997    1.9823
     param_2   -2.0143   -0.6663    0.0006    0.6613    1.9282
     param_3   -1.9686   -0.6567    0.0211    0.7019    1.9662
     param_4   -1.9411   -0.6457    0.0351    0.6831    1.9646
     param_5   -1.9145   -0.6965   -0.0176    0.6522    1.8812
```


src/chains.jl
```diff
@@ -333,6 +333,78 @@ Base.last(c::Chains) = last(c.value[Axis{:iter}].val)
...
 
+"""
+    start_time(c::Chains)
+
+Retrieve a vector of start times (as `DateTime`)
+from `chain.info`. Assumes that either
+`DateTime` or unix timestamps (`Float64`) are stored in
+`chain.info.start_time`
+"""
+function start_time(c::Chains)
```
devmotion, line 344: Maybe one should just use a special `struct` and dispatch on it?

cpfiffer: I changed the internals here to dispatch on the type of `c.info.start_time` and `c.info.stop_time`.

src/chains.jl
```diff
@@ -333,6 +333,94 @@ Base.last(c::Chains) = last(c.value[Axis{:iter}].val)
...
+# Convenience functions to handle different types of 
+# timestamps.
+to_datetime_vec(ts::DateTime) = [ts]
+to_datetime_vec(ts::Vector{DateTime}) = ts
+to_datetime_vec(ts::Float64) = [unix2datetime(ts)]
+to_datetime_vec(ts::Vector{Float64}) = unix2datetime.(ts)
+function to_datetime_vec(ts)
+    @warn "Timestamp type $(typeof(ts)) unknown."
+    return missing
+end
```
devmotion, line 345: It seems you are always only interested in `min_start` and `max_stop`, so it seems a bit inefficient to create all these vectors. Maybe just define
```suggestion
# Convenience functions to handle different types of 
# timestamps.
min_datetime(t::DateTime) = t
min_datetime(ts::Vector{DateTime}) = minimum(ts)
min_datetime(t::Float64) = unix2datetime(t)
min_datetime(ts::Vector{Float64}) = unix2datetime(minimum(ts))
min_datetime(ts) = missing_datetime(typeof(ts))

max_datetime(t::DateTime) = t
max_datetime(ts::Vector{DateTime}) = maximum(ts)
max_datetime(t::Float64) = unix2datetime(t)
max_datetime(ts::Vector{Float64}) = unix2datetime(maximum(ts))
max_datetime(ts) = missing_datetime(typeof(ts))

# does not specialize on `typeof(T)`
function missing_datetime(T::Type)
    @warn "timestamp of type $(T) unknown"
    return missing
end
```

src/chains.jl
```diff
@@ -333,6 +333,94 @@ Base.last(c::Chains) = last(c.value[Axis{:iter}].val)
...
+
+function max_stop(c::Chains)
+    ts = stop_time(c)
+
+    if ts === missing
+        return missing
+    else
+        return maximum(ts)
+    end
+end
```
devmotion, line 403: It seems one only needs `min_start` and `max_stop`. Therefore I suggest, in addition to the suggestions above:

```suggestion
"""
    min_start(c::Chains)

Retrieve the minimum of the start times (as `DateTime`) from `chain.info`.

It is assumed that the start times are stored in `chain.info.start_time` as
`DateTime` or unix timestamps of type `Float64`.
"""
function min_start(c::Chains)
    return if :start_time in keys(c.info)
        # We've got some times, return the minimum.
        min_datetime(c.info.start_time)
    else
        # Times not found -- spit out missing.
        missing
    end
end

"""
    max_stop(c::Chains)

Retrieve the maximum of the stop times (as `DateTime`) from `chain.info`.

It is assumed that the start times are stored in `chain.info.stop_time` as
`DateTime` or unix timestamps of type `Float64`.
"""
function max_stop(c::Chains)
    return if :stop_time in keys(c.info)
        # We've got some times, return the minimum.
        max_datetime(c.info.stop_time)
    else
        # Times not found -- spit out missing.
        missing
    end
end
```

src/chains.jl
```diff
@@ -333,6 +333,94 @@ Base.last(c::Chains) = last(c.value[Axis{:iter}].val)
...
+    latest_stop = max_stop(c)
+
+    # DateTime - DateTime returns a Millisecond value,
+    # divide by 1k to get seconds.
+    if earliest_start === missing || latest_stop === missing
+        return missing
+    else
+        return Dates.value(latest_stop - earliest_start) / 1000
+    end
+end
```
devmotion, line 423: I suggest that one adds an option to provide the earliest start and the latest stop, similar to e.g. `mean` in `var`, to avoid repeated computations below:
```suggestion
"""
    wall_duration(c::Chains; start=min_start(c), stop=max_stop(c))

Calculate the wall clock time for all chains in seconds.

The duration is calculated as `stop - start`, where as default `stop`
is the latest stopping time and `start` is the earliest starting time.
"""
function wall_duration(c::Chains; start=min_start(c), stop=max_stop(c))
    # DateTime - DateTime returns a Millisecond value,
    # divide by 1k to get seconds.
    return if start === missing || stop === missing
        missing
    else
        Dates.value(stop - start) / 1000
    end
end
```

src/chains.jl
```diff
@@ -450,6 +538,11 @@ function header(c::Chains; section=missing)
         "= $(join(map(string, arr), ", "))\n"
     )
 
+    # Get the wall time
+    start = minimum(start_time(c))
+    stop = maximum(stop_time(c))
+    wall = wall_duration(c)
```
devmotion, line 544: With the suggestions above one can write:
```suggestion
    start = min_start(c)
    stop = max_stop(c)
    wall = wall_duration(c; start=start, stop=stop)
```

src/chains.jl
```diff
@@ -466,11 +559,12 @@ function header(c::Chains; section=missing)
         push!(section_strings, section_string)
     end
 
-    #
-
     # Return header.
     return string(
         ismissing(c.logevidence) ? "" : "Log evidence      = $(c.logevidence)\n",
+        ismissing(start) ? "" : "Start time        = $(start)\n",
+        ismissing(start) ? "" : "Stop time         = $(stop)\n",
```
devmotion, line 566:
```suggestion
        ismissing(stop) ? "" : "Stop time         = $(stop)\n",
```

src/ess.jl
```diff
@@ -211,8 +211,15 @@ function ess(
     # estimate the effective sample size and rhat
     ess, rhat = ess_rhat(_chains.value.data; kwargs...)
 
+    # Calculate ESS/minute if available
+    dur = wall_duration(chains)
+
     # convert to namedtuple
-    nt = merge((parameters = names(_chains),), (ess = ess, rhat = rhat))
+    nt = if ismissing(dur)
```
devmotion, line 218: I assume that similar to `nothing` it is sometimes better to check `=== missing` instead of `ismissing` (I guess above it does not matter since it does not affect the output type):
```suggestion
    nt = if dur === missing
```

test/diagnostic_tests.jl
```diff
@@ -58,6 +59,18 @@ chn_disc = Chains(val_disc, start = 1, thin = 2)
     @test 0.95 ≤ mean(chn, "param_1") ≤ 1.05
 end
 
+@testset "Chain times" begin
+    t1 = time()
+    t2 = t1 + 1.5
+    chn_timed = Chains(val, info = (start_time=t1, stop_time=t2))
+
+    @test MCMCChains.start_time(chn_timed) == [unix2datetime(t1)]
+    @test MCMCChains.stop_time(chn_timed) == [unix2datetime(t2)]
```
devmotion, line 68:
```suggestion
```

devmotion (changes requested):
I left some comments. Could we add more tests and in particular test all the different dispatches, probably with `Float64`s and `DateTime`s directly? Additionally, it would be good to check that all methods that are affected by this change can be inferred correctly.

devmotion (approved):

PR #291, opened 2021-04-22
Revert update to PrettyTables 0.12

devmotion:
The dependency was not tested: https://github.com/TuringLang/MCMCChains.jl/pull/286#issuecomment-824809937


devmotion:
The main motivation for reverting it is that PrettyTables broke stuff in MCMCChains before and that it might end up untested in some release if the compatibilities are incorrect in the master branch.

andreasnoack:
Based on the discussion in https://github.com/TuringLang/MCMCChains.jl/pull/294, I believe this one can now be closed.

PR #292, opened 2021-04-23
Bump PrettyTables to version 1 in package project file and DataFrames

andreasnoack:
to version 1 in test project file. This should hopefully supersede #288, #289, #290, and #291. MLJ is now compatible with PrettyTables version 1 so there is a chance that this PR will pick up the latest versions.


devmotion:
But if the latest versions are tested, PrettyTables 0.12 would still be untested,  it seems?

andreasnoack:
Fair point. I've restarted the tests in https://github.com/TuringLang/MCMCChains.jl/pull/290. That should test with 0.12 and then we continue with this one afterward.

andreasnoack:
Following up on the discussion in https://github.com/TuringLang/MCMCChains.jl/pull/294. I've modified this PR to now allow PrettyTables 0.12. We cannot test that version since MLJ skipped it and went directly from 0.11 to 1.0.

andreasnoack:
I've added 0.12 back since the conclusion in #294 was the 0.12 had been tested on the runs that exclude MLJ testing.

andreasnoack:
Tests are passing and version 1 of both DataFrames and PrettyTables are being used. Yay! @devmotion could you please verify and merge if you agree? Once this one is in, I guess we should "rebase" https://github.com/TuringLang/MCMCChains.jl/pull/293. Then a new tag shouldn't be far away and we can move on to the next package that needs to be made DataFrames@1 ready.

devmotion (approved):
Looks good, thank you for pushing this update!

devmotion:
#293 is already included here, isn't it? So it seems we can just release 4.9.0 once this PR is merged.

andreasnoack:
You are right. I mentioned exactly that in the top post but had forgotten in the meantime.

> ...we can just release 4.9.0 once this PR is merged.

Is the policy here to bump the minor version with compact changes. IMO it should be safe just to bump the patch version. (Not that it matters too much.)

devmotion:
> Is the policy here to bump the minor version with compact changes. IMO it should be safe just to bump the patch version. (Not that it matters too much.)

There's no strict policy but we tend to bump the minor version (IIRC the reasoning is mentioned in ColPrac: compatibility with new versions of the dependencies is a new feature, not a bugfix).

devmotion:
I updated the version number and made a new release.

PR #303, opened 2021-05-07
Add compute time instead of wall time

cpfiffer:
Wall time isn't super good for cases where we have multiple chains sampled in parallel, so I've switched over to computing `compute_time` by default. `compute_time` is simple the total seconds across all chains.

In addition, I fixed what happens to start and stop times when chains are merged. Previously other chain times were being discarded, and now they are being merged accordingly. There's also some minor style fixes in here.


cpfiffer:
The error seems to be unrelated to this -- looks to be a doctest failing because of a deprecation warning in AxisArrays.

scratch.jl
```diff
@@ -0,0 +1,12 @@
+using Revise
```
yebai, line 1: Is this file meant to be tracked?

yebai (approved):

src/chains.jl
```diff
@@ -748,11 +775,18 @@ function _cat(::Val{3}, c1::Chains, args::Chains...)
     all(c -> names(c) == nms, args) || throw(ArgumentError("chain names differ"))
 
     # concatenate all chains
-    data = mapreduce(c -> c.value.data, (x, y) -> cat(x, y; dims = 3), args;
-                     init = c1.value.data)
+    combined = [c1, args...]
```
devmotion, line 778: The current implementation deliberately avoids splatting of `args`. Why is this needed? Could we also check if everything is still inferred correctly?

src/chains.jl
```diff
@@ -548,43 +574,44 @@ function header(c::Chains; section=missing)
     # Return header.
     return string(
         ismissing(c.logevidence) ? "" : "Log evidence      = $(c.logevidence)\n",
-        ismissing(start) ? "" : "Start time        = $(start)\n",
-        ismissing(stop) ? "" : "Stop time         = $(stop)\n",
-        ismissing(wall) ? "" : "Wall duration     = $(round(wall, digits=2)) seconds\n",
         "Iterations        = $(first(c)):$(last(c))\n",
         "Thinning interval = $(step(c))\n",
-        "Chains            = $(join(map(string, chains(c)), ", "))\n",
+        "Chains            = $(length(map(string, chains(c))))\n",
```
devmotion, line 579: Is it correct that with this change only the number of chains will be displayed? If this is the case, maybe change it to
```suggestion
        "Number of chains  = $(size(c, 3))\n",
```
now that only the number of chains is displayed?

Issue #246, opened 2020-10-20
[RFC] MCMCChains Goals


cpfiffer:
I wanted to collate some of the longer-term goals for MCMCChains. I think many of us would agree that MCMCChains is quite stable at this point. It generally works well, has an acceptably defined set of indexing and analysis tools, and has lots of interactions with the rest of the ecosystem (not just Turing anymore!).

However, I do think it's worth putting down in words the various issues we still have with MCMCChains. This issue will be the hub for general improvements to the package, big or small, that we hope we can get some other people to help out on. 

Comments are appreciated.

Improvements:

## Performance, particularly for `display` calls. 

The first time the user calls display (explicitly or through REPL usage), it takes my machine about 6.5 seconds just to handle to display call. If you're a REPL user this is absolutely brutal. Granted, `display(chain)` does a ton of numerical work and thus lights up a lot of method compilation, but I wonder if there's a way to make the user experience here slightly better if possible.

## Adding tools to reshape parameters into their original forms

We have a recent issue on this (https://github.com/TuringLang/MCMCChains.jl/issues/231) but it has appeared consistently for as long as I have been working on MCMCChains (nearly 2 years now). For example, if you have two parameters `x[1]=4` and `x[2]=2`, it's pretty easy for us to say that these could be repacked into their semantic form `x = [4, 2]` and just given to the user that way. As of right now, it's not super easy to do this kind of thing -- you typically have to roll your own code that matches your model setup. `get` kind of solves this for you, but it could use a lot more work to create the `(x=[4,2],)` named tuple I would like to see.

## Visual improvements to plotting and generally better heuristics

The plotting functionality of MCMCChains is pretty good, but it suffers from a lack of polish and doesn't have as many intuitive safeguards. For example, if you plot a chain with many parameters, by default all parameters will be thrown up with their own tiny density and trace plots that are completely unreadable. Obviously you can fix this quickly by just subsetting the chain (`plot(chain["param1"])`), but there should be a more palatable plot if the number of parameters is obscenely large. I think `rstan` has a good approach to this where it throws up just a mean and 95% posterior density for each parameter -- it's very compact. 

On another note, we don't seem to specify plot spacing, so the text on plots can get smushed into other subplots. I'm not sure if this is a Plots.jl issue or our issue.

## Heterogeneous variable types

[...]

PR #307, opened 2021-06-10
Make Chains objects display only information and not statistical eval

PaulinaMartin96:
As suggested on #246, Chains objects now display only information and not statistical evaluations.

Before:
MCMCChainsnotmodified

After:
MCMCChainsmodified

Also, the describe function now display both Summary statistics and Quantiles info.

Before:
describenotmodified

After:
describemodified


cpfiffer:
Woops, I missed this one as reviewable. If I forget, as I do often, you can flag me by requesting a review over here: 

image

cpfiffer (commented):
For PRs like this where a UI change is involved, try to include a printout of the new behavior so someone can quickly look at the PR to see what it looks like. For example, I would want to see the results of `show(chain)`, which is the primary change introduced by this PR. 

Could you add that for me so I can take a look at the end result that a user might see?

Also, we follow [ColPrac](https://github.com/SciML/ColPrac), one key feature of which is always making sure that the version number of whatever package you submit a PR to is increased according to [semantic versioning](https://semver.org/). You can find the version number for MCMCChains in `Project.toml`, [here](https://github.com/TuringLang/MCMCChains.jl/blob/46b12c55370b6ed4e309f4f742e8384789eb3eaa/Project.toml#L6). 

For a change like this where the default behavior changes, I would recommend bumping us up to MCMCChains 4.14.0.

Flag me after you add a little description of what's changing and what I should expect to see and I'll do another pass on this one.

PaulinaMartin96:
@cpfiffer thank you so much for the suggestions! I'll consider them in the future. I think the PR is ready. There's a conflict with Project.toml. Let me know if further changes are required.

cpfiffer:
Okay, thanks for adding that stuff. Looks good to me. 

When there's a merge conflict, it's typically the PR person's job to fix the merge conflict. Fortunately this one is really easy -- you can use the GitHub web editor or your local git merge manager to fix the issues. GitHub has links on how to do this at the bottom of the PR scroll:

image

This is a good learning opportunity since it's one of the easier merge conflicts. Go ahead and try to fix this issue. In this case, you basically want to fix the version number to the one you specified (4.14.0) and not the one that's currently in master.

PaulinaMartin96:
Sorry for that, and again, thank you so much for your advice!

cpfiffer:
No problem, all part of the learning process. I've just enabled tests and will check back to make sure they've passed.

src/stats.jl
```diff
@@ -171,6 +171,15 @@ function describe(
     return dfs
 end
 
+function Base.show(io::IO, mime::MIME"text/plain", cs::Vector{ChainDataFrame})
```
devmotion, line 174: This only captures the abstract `Vector{ChainDataFrame}` but not any concretely typed `Vector{<:ChainDataFrame}`. In general, I am a bit worried about changing the display of vectors of `ChainDataFrame`s - it seems wrong to completely opt out of the default display mechanism of vectors in Julia (I also wonder if it causes any problems) just to change the way in which `describe(chain)` is displayed. Maybe rather `describe` should not return a vector of `ChainDataFrame`.

cpfiffer: Yeah, I'm down with this. My understanding of your suggestion is that `describe` should be a pure IO function -- we should make `display(io::IO, chn::Chain)` and the output is all the stuff inside the `Base.show` definition above.

devmotion: We should not define `display` (it calls `show(io, MIME("text/plain"), x)` which should be implemented), but we should just implement `DataAPI.describe(io, chain)` if we want to display the summary statistics in a nice way: https://juliastats.org/StatsBase.jl/latest/scalarstats/#Summary-Statistics-1

PaulinaMartin96: Thank you so much for your help! So, to double check that I'm understanding these suggestions. Instead of returning a `Vector{ChainDataFrame}`,  `describe` should be an implementation of `StatsBase.describe` and return something like this?

```julia
chn = Chains(rand(100, 2, 2), [:a, :b])
chn_arr = Array(chn)
sections = chn.name_map[:parameters]
for i in 1:length(sections)
    println("Parameter $(sections[i])")
    describe(chn_arr[:,i])
    println()
end
```
StatsBase _describe

devmotion: I think it's fine if `summarystats` returns different summary statistics (so this method does not have to be changed) but we should make sure that `describe` just prints these summary statistics in a pretty way to be consistent with how `describe` and `summarystats` are defined in StatsBase. I.e., in particular `describe` should not return anything but only print to IO and it should not print the quantiles if they are not part of `summarystats` (here it might actually be better to include them in `summarystats` as well and return two dataframes, possibly as a named tuple).

cpfiffer: To clarify maybe a little bit, there's no need to change to this code you've posted:

```julia
chn = Chains(rand(100, 2, 2), [:a, :b])
chn_arr = Array(chn)
sections = chn.name_map[:parameters]
for i in 1:length(sections)
    println("Parameter $(sections[i])")
    describe(chn_arr[:,i]) # <- This is not what we want, we want to print the results of `describe(chain)` here instead
    println()
end
```

Basically we want to change the stuff that is _currently_ in `show` to `describe`, so that `describe` becomes a pure IO function and not a weird `Vector{<:ChainDataFrame}` thing that we have now.

Maybe one way is to do something like

```julia
function DataAPI.describe(io::IO, chains::Chains)
    print(io, "Chains ", chains, ":\n\n", header(chains))

    summstats = summarystats(chains)
    qs = quantiles(chains)

    println(io)
    show(io, summstats)

    println(io)
    show(io, qs)
end
```

which won't actually return anything, it just prints stuff out to the screen. There's probably a way more sane way to do this, but it's a rough sketch to get you started.

shravanngoswamii:
Can someone please review this PR, I have made these changes:
- Implemented DataAPI.describe()
- display() only prints the header now, with a suggestion to use describe()

See:
```Julia
julia> using MCMCChains

julia> chn = Chains(rand(100, 2, 2), [:a, :b])
Chains MCMC chain (100×2×2 Array{Float64, 3}):

Iterations        = 1:1:100
Number of chains  = 2
Samples per chain = 100
parameters        = a, b

Use `describe(chains)` for summary statistics and quantiles.


julia> display(chn)
Chains MCMC chain (100×2×2 Array{Float64, 3}):

Iterations        = 1:1:100
Number of chains  = 2
Samples per chain = 100
parameters        = a, b

Use `describe(chains)` for summary statistics and quantiles.


julia> describe(chn)
Chains MCMC chain (100×2×2 Array{Float64, 3}):

Iterations        = 1:1:100
Number of chains  = 2
Samples per chain = 100
parameters        = a, b

Summary Statistics
  parameters      mean       std      mcse   ess_bulk   ess_tail      rhat   ess_per_sec 
      Symbol   Float64   Float64   Float64    Float64    Float64   Float64       Missing 

           a    0.4691    0.2997    0.0210   181.4680   106.1712    1.0190       missing
           b    0.4721    0.3108    0.0223   205.3635   184.7170    1.0254       missing

Quantiles
  parameters      2.5%     25.0%     50.0%     75.0%     97.5% 
      Symbol   Float64   Float64   Float64   Float64   Float64 

           a    0.0123    0.1906    0.4581    0.7287    0.9712
           b    0.0217    0.1789    0.4507    0.7560    0.9812

julia> 
```

cc @yebai

penelopeysm (commented):
Isn't this a breaking change to the behaviour of `describe`? Does this need a major version bump?

shravanngoswamii:
> Isn't this a breaking change to the behaviour of `describe`? Does this need a major version bump?

Yep, It does need a version bump! I just committed the version bump from 6.0.7 to 6.1.0

penelopeysm (approved):
That looks good to me!

devmotion:
> The format workflow is failing due to permission issues since this PR is from a fork.

Only posting of the suggested changes fails. The action would succeed if they code would be formatted correctly.

shravanngoswamii:
Thank you @PaulinaMartin96 @cpfiffer @devmotion @penelopeysm @yebai!

shravanngoswamii:
@penelopeysm How can I trigger a release for v7.0.0, just by creating a tag or any other steps?

PR #315, opened 2021-07-04
Document mcse

rikhuijzer:


devmotion:
The function is/will be moved to InferenceDiagnostics, so the docstring there should be updated (as well).

rikhuijzer:
PR opened at https://github.com/devmotion/InferenceDiagnostics.jl/pull/16.

PR #319, opened 2021-07-19
Prior/posterior predictive check plots

PaulinaMartin96:


PaulinaMartin96:
For this PR, should the version be 4.16.0 (after #316 ) or 5.1.0 (after #310 )?

cpfiffer:
Probably 4.16.0 since #310 is a bigger thing and probably won't have too much effect here.

yebai:
Looks very useful. Thanks @shravanngoswamii. 

@sunxd3, you might be interested in this, too.

penelopeysm (approved):
Happy with this. Realistically I don't think anybody is going to be constructing Chains manually and plotting them so it would probably be a good thing to have some docs focusing on how to generate the data with Turing, but that's not a part of this PR

PR #322, opened 2021-07-28
Bump AxisArrays version (Fixes Julia 1.7)

rikhuijzer:
AxisArrays works now on Julia 1.7 thanks to https://github.com/JuliaArrays/AxisArrays.jl/pull/201.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -31,7 +31,7 @@ Tables = "bd369af6-aec1-5ad0-b16a-f7cc5008161c"
 [compat]
 AbstractFFTs = "0.5, 1"
 AbstractMCMC = "0.4, 0.5, 1.0, 2.0, 3.0"
-AxisArrays = "0.4.2"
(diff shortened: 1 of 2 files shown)


rikhuijzer:
The single failure on nightly x64 (Ubuntu, Windows, MacOS) is caused by a PR from a few days ago (https://github.com/JuliaLang/julia/pull/41619) and will be fixed in https://github.com/JuliaStats/StatsBase.jl/pull/703/. The failure on nightly x86 seems like a bug.

rikhuijzer:
The explicit pinning of AxisArrays to 0.4.4 is due to https://github.com/TuringLang/MCMCChains.jl/pull/213. Since AxisArrays 0.4.4 works on Julia 1.0 as well as Julia 1.7, I don't think it is necessary to set AxisArrays = "0.4.2, 0.4.4". But, of course, there are also arguments to use both.

PR #323, opened 2021-07-30
Add ridgeline plot

PaulinaMartin96:
This PR includes three plot types: Ridgeline, forest and caterpillar plots. These recipes functions receive

**args**
- Chains object
-  par_names: a vector of `Symbol`. This can be a parameter section, a subset of parameters, or only one parameter.

**kwargs**
- `hpd_val` = [0.05, 0.2]. A n-element vector to calculate the desired HPD intervals. Values are interpreted as in the MCMCChains hdp function. For ridgeline plots, only the smaller value will be taken. For caterpillar and forest plots, all the values in the vector will be used. 
- `q` = [0.1, 0.9]. A two-element vector to calculate quantiles intervals.
- `spacer` = 0.4.  Used for determining spacing between plots. This should be modified to adjust
- `riser` = 0.2. Used for determining spacing between plots. 
- `show_mean` = true,
- `show_median` = true,
- `show_qi` = true
- `show_hpdi` = false,
- `fill_q` = false,
- `fill_hpd` = true
- `ordered` = false

In order to compute the required data for the plot, an auxiliary function `_compute_plot_data` was defined. As `args`, this function receives a Chains object (`chains`), a vector of `Symbol`  (`par_names`) and a `plot_type` (`:ridge`, `:caterpillar`, `:forest`), and as `kwargs` it receives the ones defined above.
 
Plotting options for each curve are presented below:

**Ridgeline Plot**

**(a) Fill**
Fill area below the curve can be determined by quantiles interval (`fill_q = true`) or hdpi interval (`fill_hpd = true`). If both `fill_q = false` and  `fill_hpd = false`, then the whole area below the curve will be filled. If no fill color is desired it should be specified with series attributes. These options are mutually exclusive.

**(b) Mean and median**
 A vertical line can be plotted repesenting the mean (`show_mean = true`) or median (`show_median = true`) of the kde distribution. Both options can be plotted at the same time. 

** (c) Intervals **
At the bottom of each density plot, a quantile interval ('show_qi = true`) or HPD interval (`show_hdpi = true`) can be plotted. These options are mutually exclusive. To plot quantile intervals, values in `q` will be taken, and for HPD intervals, only the smaller value specified in `hpd_val` will be used.

_Note:_ When one parameters is given, it will be plotted as a density plot with all the elements described above.

**Forest and Caterpillar plots**

Both plot type are called using `forestplot` shorthands. If `ordered = false` then a forest plot will be returned, and on the contrary, if `ordered = true` a caterpillar plot will be returned. 

For both plot types the following elemets can be plotted:

-  **HPDI** determined by the number of element of the `hpd_val` kwarg. All the values will be used to construct the intervals with `MCMCChains.hpd`. 
- **Quantile interval** determined by the 2-element vector `q`
-**Mean** and/or **median**. Plotted as points with different `markershape`.

Working example:

```julia
using MCMCChains, StatsPlots, Statistics

n_iter = 100
n_name = 3
n_chain = 2

val = randn(n_iter, n_name, n_chain) .+ [1, 5, 10]'
val = hcat(val, rand(11:12, n_iter, 1, n_chain))

chn = Chains(val, [:A, :B, :C, :D])
val2 = rand(n_iter, 11, n_chain) .+ ([1, 2, 3, 4, 5, 6, 5.5, 4.5, 3.5, 2.5, 1.5]' .* rand())
chn2 = Chains(val2) 
```

```julia
p1 = ridgelineplot(chn, chn.name_map[:parameters], fill_q = true, fill_hpd = false, legend = :topleft)
p2 = ridgelineplot(chn, chn.name_map[:parameters], fill_q = false, fill_hpd = true, legend = :topleft)
p3 = ridgelineplot(chn, chn.name_map[:parameters], fill_q = false, fill_hpd = false, legend = :topleft)
p4 = ridgelineplot(chn, chn.name_map[:parameters], fill_q = false, fill_hpd = false, fillcolor = :white, legend = :topleft)
p5 = ridgelineplot(chn, [:A, :B, :C, :D], palette = :summer, legend = :topleft)
p6 = ridgelineplot(chn, [:A, :B, :C], show_hpdi = false, show_qi = true, palette = :Blues_5, legend = :topleft)
p7 = ridgelineplot(chn, [:A, :B], show_hpdi = false, show_qi = false, legend = :topleft)
p8 = ridgelineplot(chn, [:A], show_median = false, show_mean = false, show_hpdi = false, show_qi = true, legend = :topleft)
plot(p1, p2, p3, p4, p5, p6, p7,p8, layout = (2,4), size = (1400, 600))
```
ridgeline

```julia
p9 = forestplot(chn2, chn2.name_map[:parameters], hpd_val = [0.05, 0.15, 0.25])
p10 = forestplot(chn2, chn2.name_map[:parameters], hpd_val = [0.05, 0.15, 0.25], ordered = true)
plot(p9, p10, layout = (1,2), size = (900, 400), legend = :topleft)
```
forest


PaulinaMartin96:
For this one, there are some comments:

1. In order to displace the plots, I had to use kde and pdf funcions from KernelDensity.jl. I was thinking of adding the info for KernelDensity found in Project.toml from test (that would be KernelDensity = "5ab0869b-81aa-558d-bb23-cbf5423bbe9b" for deps and KernelDensity = "0.6.2" for compat). Let me know if this change can be made.

2. The recipes for these plot type are user recipes. There is only a shorthands version for calling 
 the plots (`ridgelineplot(chn, par_names; kwargs...)` and `forestplot(chn, par_names; kwargs...)` ) but no a "standard version" (`plot(args...; seriestype = ridgelineplot)`). This happens because of argument overlapping, so I'll fix this later. I tried to implement these plots with other recipes types (type and plot recipes) but user recipes were the best option.

3. Ridgelineplots are not implemented for histograms. I tried various options but none of those were effective. There is one option that plays with layout (with inset and bbox function), but it's necessary to add Plots as a dependency.

cpfiffer:
1. Yeah, I'm fine with adding KernelDensity to the dependencies.
2. Sounds good to me!
3. Let's keep `Plots` out of our dependencies, since it's kind of a massive import that we don't want to bring along.

src/plot.jl
```diff
@@ -184,3 +186,214 @@ end
...
+    _riser = 0.2,
+    barbounds = (-Inf, Inf),
+    show_mean = true,
+    show_median = true,
+    show_qi = false,
+    show_hpdi = true,
+    fill_q = true,
+    fill_hpd = false,
+    ordered = false
+    )
```
cpfiffer, line 209:
```suggestion
)
```

src/plot.jl
```diff
@@ -184,3 +186,214 @@ end
...
+    spacer = 0.5,
+    _riser = 0.2,
+    show_mean = true,
+    show_median = true,
+    show_qi = false,
+    show_hpdi = true,
+    fill_q = true,
+    fill_hpd = false,
+    ordered = false
+    )
```
cpfiffer, line 254:
```suggestion
)
```

src/plot.jl
```diff
@@ -184,3 +186,214 @@ end
...
```
cpfiffer, line 341:
```suggestion
)
```

src/plot.jl
```diff
@@ -184,3 +186,214 @@ end
...
+    end
+    chain_med = median(chain_vec)
+    chain_mean = mean(chain_vec)
+    min = minimum(k_density.density .+ h)
+    q_int = (show_qi ? [qs[1], chain_med, qs[2]] : [chain_med])
+    return _plot_data(par, hpdi, lower_hpd, upper_hpd, h, qs, k_density,
+        x_int, val, chain_med, chain_mean, min, q_int)
+end
+
+@recipe function f(p::RidgelinePlot;
```
cpfiffer, line 242:
```suggestion
@recipe function f(
    p::RidgelinePlot;
```

src/plot.jl
```diff
@@ -184,3 +186,214 @@ end
...
+            label := (i == 1 ? "$(Integer((1-hpdi[1])*100))% HPDI" : nothing)
+            linewidth --> (show_hpdi ? 2 : 0)
+            seriesalpha --> 0.80
+            linecolor --> :darkblue
+            [lower_hpd[1][1], upper_hpd[1][1]], [h, h]
+        end
+    end
+end
+
+@recipe function f(p::ForestPlot;
```
cpfiffer, line 329:
```suggestion
@recipe function f(
    p::ForestPlot;
```

cpfiffer (approved):
Looks good to me. Could you push the updated Project.toml file too with an incremented version number?

cpfiffer (approved):

cpfiffer:
Should be alright to merge if tests pass.

PaulinaMartin96:
I think it's ready now.  By adding KernelDensity, a Warning message appears about the function `getproperty()`. This functions is defined in Tables.jl ([lines 7-9](https://github.com/JuliaData/Tables.jl/blob/main/src/Tables.jl)) and in Compat.jl from ChainRulesCore ([lines 1-3](https://github.com/JuliaDiff/ChainRulesCore.jl/blob/master/src/compat.jl); only for versions less than v"1.2"). When adding KernelDensity this function is overwritten. So far I haven't found any problem on v1.1.1;  v.1.6.0 has no warning message and works fine.  

warning_getproperty

cpfiffer:
Yeah, that's a `ChainRulesCore` thing -- not really our problem.

cpfiffer:
Let's goooooooooooo

Project.toml
```diff
@@ -3,7 +3,7 @@ uuid = "c7f686f2-ff18-58e9-bc7b-31028e88f75d"
 keywords = ["markov chain monte carlo", "probablistic programming"]
 license = "MIT"
 desc = "Chain types and utility functions for MCMC simulations."
-version = "4.13.2"
+version = "4.17.0"
```
rikhuijzer, line 6: Why is this version bumped so much and why isn't it registered? @cpfiffer

cpfiffer: Woops an oversight on my part. The overbump was because there's other PRs that needed to be merged and it wasn't clear which was first. I'll fix and tag this.

storopoli:
Nice addition @PaulinaMartin96. I was trying the new plots with the example you provided and I cannot make a catterpilar plot (`ordered=true`), it just returns a regular forestplot.

PaulinaMartin96:
Thank you so much for comment @storopoli! I checked what you said and you're right. It's a problem with the `kwargs`. I'll fix it right away.

PR #341, opened 2021-12-22
Remove direct dependency on SpecialFunctions

devmotion:
It seems SpecialFunctions is not used anymore. It was used in e.g. the Geweke diagnostic but these parts of MCMCChains were moved to MCMCDiagnosticTools (which still depends on SpecialFunctions).

Fixes https://github.com/TuringLang/MCMCChains.jl/pull/338.


rikhuijzer (approved):

devmotion:
Don't merge, doctests are broken due to different printing on arrays on Julia 1.7 and hence the documentation is not built. I'll open a separate PR.

PR #342, opened 2021-12-23
Fix doctests

devmotion:


devmotion:
Weird, I ran `doctest(MCMCChains; fix=true)` locally with Julia 1.7 and everything seems to pass.

rikhuijzer:
How about throwing away some compatibility stuff for Julia version 1.6 and lower? We're helping almost no one with that and just making the lives of the few Turing developers more difficult

devmotion:
The main problem is not 1.0. We don't have to do anything special in this package to support it, so I don't think we should drop it.

The main problems are 1) that array printing is different depending on the Julia version, also from 1.6 to 1.6, and 2) the MLJ dependencies for the Rstar tests. Both require fixes even if we would support only >= 1.6.

The XGBoost dependency is annoying (and eg doesn't work on 32bit so we can't run tests there), and hence probably it would be helpful to use some simpler model in the tests (eg in MCMCDiagnosticTools I use Libsvm as well but it also has binary dependencies, so maybe eg naive Bayes would be even better). Additionally, there's a new release of MLJBase that breaks our tests - however, fixing it to an old version breaks our current test setup.

rikhuijzer:
> The main problem is not 1.0. We don't have to do anything special in this package to support it, so I don't think we should drop it.

This is special:

https://github.com/TuringLang/MCMCChains.jl/blob/e7b3db5a4bb09e0c7dcdffc82a51461e084faa54/test/runtests.jl#L3-L8

https://github.com/TuringLang/MCMCChains.jl/blob/e7b3db5a4bb09e0c7dcdffc82a51461e084faa54/test/runtests.jl#L22-L27

rikhuijzer:
> and eg doesn't work on 32bit so we can't run tests there

Also, who runs 32 bit in this day and age? Apple dropped support for 32-bit in OS X Lion in 2011.

rikhuijzer:
Also, SciML drops all support for versions lower than Julia 1.6: https://discourse.julialang.org/t/scimls-stance-on-the-new-julia-lts/72446.

devmotion:
These specialization are not the problem here, and at least the second one is still needed even if we drop < 1.6.

32bit tests helped to uncover Int32/Int64 issues.

rikhuijzer:
> 32bit tests helped to uncover Int32/Int64 issues.

nice okay

devmotion:
I fixed the issues and the annoying binary dependencies by replacing XGBoost with DecisionTree.

There are no additional fixes or workarounds for specific Julia versions needed, and actually now the rstar statistic can be tested with any Julia version and also on 32bit architectures

Doctests are only performed with Julia >= 1.7 and on 64bit which fixes all issues with array printing in different Julia versions (again, not a Julia 1.0 thing but also a problem for Julia 1.6 vs 1.7). IMO doctests really should only be performed with one specific Julia versions due to changes in e.g. `print`, `show`, and RNGs (doctest filters can mask outdated documentation since matches in the expected and actual outcome will be removed before comparing them). However, using `>= v"1.7"` instead of `== v"1.7"` ensures that we don't silently drop doctests when a new Julia version is released and CI is updated.

test/runtests.jl
```diff
@@ -96,4 +68,19 @@ Random.seed!(0)
...
+    # therefore we only run doctests with Julia >= 1.7 and on 64bit where `Int === Int64`
+    if VERSION >= v"1.7" && Sys.WORD_SIZE == 64
+        DocMeta.setdocmeta!(
+            MCMCChains,
+            :DocTestSetup,
+            :(using MCMCChains);
+            recursive = true
+        )
+        doctest(MCMCChains)
+    end
```
rikhuijzer, line 85: Awesome. Great improvement

rikhuijzer (approved):
Looks very good to me. Pure Julia is definitely better than XGBoost with its binary dependency

PR #345, opened 2021-12-24
Add Makie.jl example

rikhuijzer:
Reasonably simple Makie.jl example. I aimed to write down some code to help people get started and allow them to fine-tune it to their use-case.

Currently, the plot looks like

image

### DIFF
--- a/docs/src/gadfly.md
+++ b/docs/src/gadfly.md
@@ -9,15 +9,6 @@ using Gadfly
 write_svg(path, p; w=6inch, h=4inch) = Gadfly.draw(Gadfly.SVG(path, w, h), p) # hide
 using MCMCChains
 
-# Define the experiment.
-n_iter = 100
-n_name = 3
-n_chain = 2
-
-# Experiment results.
-val = randn(n_iter, n_name, n_chain) .+ [1, 2, 3]'
-val = hcat(val, rand(1:2, n_iter, 1, n_chain))
-
 chn = Chains(randn(100, 2, 3), [:A, :B])
 df = DataFrame(chn)
 df[!, :chain] = categorical(df.chain)
--- a/docs/src/makie.md
+++ b/docs/src/makie.md
@@ -5,7 +5,6 @@ The example is meant to provide an useful basis to build upon.
 Let's define some random chain and load the required packages:
 
 ```@example makie
-using CairoMakie
 using MCMCChains
 
 chns = Chains(randn(300, 5, 3), [:A, :B, :C, :D, :E])
@@ -15,6 +14,9 @@ A basic way to visualize the chains is to show the drawn samples at each iterati
 Colors depict different chains:
 
 ```@example makie
+using CairoMakie
+CairoMakie.activate!(type = "svg")
+
 params = names(chns, :parameters)
 
 n_chains = length(chains(chns))
--- a/docs/src/makie.md
+++ b/docs/src/makie.md
@@ -6,7 +6,6 @@ Let's define some random chain and load the required packages:
 
 ```@example makie
 using CairoMakie
-using DataFrames
 using MCMCChains
 
 chns = Chains(randn(300, 5, 3), [:A, :B, :C, :D, :E])
@@ -18,22 +17,21 @@ Colors depict different chains:
 ```@example makie
 params = names(chns, :parameters)
 
-df = DataFrame(chns)
-n_chains = length(unique(df.chain))
-n_samples = nrow(df) / n_chains
-
-# Alternatively, use `CategoricalArrays.categorical`.
-df[!, :chain] = string.(df.chain)
+n_chains = length(chains(chns))
+n_samples = length(chns)
 
 fig = Figure(; resolution=(1_000, 800))
 
+function chain_values(chns, chain::Int, param::Symbol)
+    return get(chns, param)[param][:, chain]
+end
+
 # Create and store separate axes for showing iterations.
 values_axs = [Axis(fig[i, 1]; ylabel=string(c)) for (i, c) in enumerate(params)]
-for (ax, col) in zip(values_axs, params)
-    for i in 1:n_chains
-        chain = string(i)
-        values = filter(:chain => ==(chain), df)[:, col]
-        lines!(ax, 1:n_samples, values; label=chain)
+for (ax, param) in zip(values_axs, params)
+    for chain in 1:n_chains
+        values = chain_values(chns, chain, param)
+        lines!(ax, 1:n_samples, values; label=string(chain))
     end
 end
 
@@ -49,11 +47,10 @@ Next, we can add a second row of plots next to it which show the density estimat
 
 ```@example makie
 density_axs = [Axis(fig[i, 2]; ylabel=string(c)) for (i, c) in enumerate(params)]
-for (ax, col) in zip(density_axs, params)
-    for i in 1:n_chains
-        chain = string(i)
-        values = filter(:chain => ==(chain), df)[:, col]
-        density!(ax, values; label=chain)
+for (ax, param) in zip(density_axs, params)
+    for chain in 1:n_chains
+        values = chain_values(chns, chain, param)
+        density!(ax, values; label=string(chain))
     end
 end
 
@@ -70,7 +67,7 @@ fig
 ```
 
 Finally, let's add a simple legend.
-Thanks to setting `label` above, this legend will have the right entries:
+Thanks to setting `label` above, this legend will have the right labels:
 
 ```@example makie
 axislegend(first(density_axs))
--- a/docs/src/makie.md
+++ b/docs/src/makie.md
@@ -1,7 +1,7 @@
 # Makie.jl plots
 
 Compared to Gadfly.jl and StatsPlots.jl, Makie.jl is the most flexible plotting library that you can use.
-On this page is an example implementation for plotting with Makie.jl which you can use to further tweak to your needs.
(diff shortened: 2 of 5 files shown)


docs/src/gadfly.md
```diff
@@ -9,15 +9,6 @@ using Gadfly
...
 
-# Define the experiment.
-n_iter = 100
-n_name = 3
-n_chain = 2
-
-# Experiment results.
-val = randn(n_iter, n_name, n_chain) .+ [1, 2, 3]'
-val = hcat(val, rand(1:2, n_iter, 1, n_chain))
-
```
rikhuijzer, line 20: This code was not used

docs/src/makie.md
```diff
@@ -0,0 +1,66 @@
...
+using MCMCChains
+
+chns = Chains(randn(300, 5, 3), [:A, :B, :C, :D, :E])
+```
+
+```@example makie
+function plot_chains(chns; density_func=density!)
+    params = names(chns, :parameters)
+
+    df = DataFrame(chns)
```
rikhuijzer, line 18: I guess this can be avoided. I'm more familiar with the DataFrames data manipulations than directly handling MCMCChains, so that is why I do this and think that it's reasonable for this example

storopoli, line 18: Can we use the `Chains` object directly instead of converting to a `DataFrame`.
I think this is possible because `MCMCChains.jl` is a `Tables.jl` API-friendly interface.

rikhuijzer: I know. Feel encouraged to show such an implementation

docs/src/makie.md
```diff
@@ -0,0 +1,66 @@
+# Makie.jl plots
+
+Compared to Gadfly.jl and StatsPlots.jl, Makie.jl is the most flexible plotting library that you can use.
```
devmotion, line 3: Is this actually true? One can tweak and combine StatsPlots (or probably Gadfly, not sure?) also a lot. IMO the advantages of Makie are quality of plots (CairoMakie pdfs and svgs) and interactivity (eg GLMakie and WGLMakie). Probably shouldn't be discussed here. Can we change it to something more neutral and less judging?

docs/src/makie.md
```diff
@@ -0,0 +1,66 @@
...
+```@example makie
+using CairoMakie
+using DataFrames
+using MCMCChains
+
+chns = Chains(randn(300, 5, 3), [:A, :B, :C, :D, :E])
+```
+
+```@example makie
+function plot_chains(chns; density_func=density!)
```
devmotion, line 15: Defining a function that is only called once seems unnecessarily complicated for an example. Just write it as a single script instead? And if possible, maybe split it into different steps, explain them and show the intermediate results similar to the MakieLayout tutorial?

docs/src/makie.md
```diff
@@ -0,0 +1,79 @@
...
+```
+
+A basic way to visualize the chains is to show the drawn samples at each iteration.
+Colors depict different chains:
+
+```@example makie
+params = names(chns, :parameters)
+
+df = DataFrame(chns)
+n_chains = length(unique(df.chain))
```
devmotion, line 22: This is just `size(chns, 3)` or the more descriptive `length(chains(chns))`.

docs/src/makie.md
```diff
@@ -0,0 +1,79 @@
...
+n_samples = nrow(df) / n_chains
```
devmotion, line 23: This is just `length(chns)`.

docs/src/makie.md
```diff
@@ -0,0 +1,79 @@
...
+
+# Alternatively, use `CategoricalArrays.categorical`.
+df[!, :chain] = string.(df.chain)
+
+fig = Figure(; resolution=(1_000, 800))
+
+# Create and store separate axes for showing iterations.
```
devmotion, line 30: Since the intermediate figure is not plotted it doesn't seem necessary to save the axes?

docs/src/makie.md
```diff
@@ -0,0 +1,79 @@
...
+values_axs = [Axis(fig[i, 1]; ylabel=string(c)) for (i, c) in enumerate(params)]
+for (ax, col) in zip(values_axs, params)
+    for i in 1:n_chains
+        chain = string(i)
+        values = filter(:chain => ==(chain), df)[:, col]
```
devmotion, line 35: One could use TableOperations.jl.

docs/src/makie.md
```diff
@@ -0,0 +1,79 @@
...
+        lines!(ax, 1:n_samples, values; label=chain)
+    end
+end
+
+# Thanks to having stored the axes before, we can apply some extra tweaks.
+# These tweaks usually depend on the kind of model being fitted.
+values_axs[end].xlabel = "Iteration"
+hidexdecorations!.(values_axs[1:end-1]; grid=false)
+
+fig
```
devmotion, line 45: Can we ouput SVGs?

docs/src/makie.md
rikhuijzer: Yes, to tweak the appearance afterwards. For example, to drop all xticks for all plots **except** the lowest one, saving allows to write `hidexdecorations!.(values_axs[1:end-1]; grid=false)`. When moving this inside the loop, that would require a check to see whether we're not in the last iteration.

Also, to add a separate Legend to the figure, having stored the axes is required.

docs/src/makie.md
rikhuijzer: I tried and couldn't get it to work.

One could also grab the right matrix row from the AxisArray :sweat_smile: This is what is implemented now
```julia
julia> get(chns, :A)[:A]
2-dimensional AxisArray{Float64,2,...} with axes:
    :iter, 1:1:300
    :chain, 1:3
And data, a 300×3 Matrix{Float64}:
  2.26218      0.214943    1.01596
 -0.428555     0.603559   -0.47663
 -0.626586    -0.819862    1.59953
[...]
```

docs/src/makie.md
rikhuijzer: Yes. Changed.

Doesn't matter much for file size though:
```
-rw-r--r-- 1 rik users 255K Dec 25 10:55 index.png
-rw-r--r-- 1 rik users 233K Dec 25 10:54 index.svg
```

rikhuijzer:
In the tutorial, I now use

```julia
function chain_values(chns, chain::Int, param::Symbol)
    return get(chns, param)[param][:, chain]
end
```

to get the values for a certain chain and param. Maybe, something like this should be an additional overload for `get`? An overload like `Base.get(c::Chains, v::Symbol, chain::Int)` maybe

docs/src/makie.md
rikhuijzer: Solved via directly pulling the right elements from the axisarrays. Those arrays store different chains in different matrix columns... :sweat_smile:

devmotion:
It seems simpler to just index into the chain.

docs/src/makie.md
devmotion: I'm not concerned about file size but the quality of the plots.

docs/src/makie.md
devmotion: Just use regular indexing, ie. `chn[:A]`? BTW one should never have to work with the underlying axis array directly,ie never use `chn.values`.

docs/src/makie.md
devmotion: > When moving this inside the loop, that would require a check to see whether we're not in the last iteration.

I know, I think it would be more natural to do this in the loop since we only plot the figure after hiding the decorations but are not interested in any intermediate figures.

> Also, to add a separate Legend to the figure, having stored the axes is required.

You can obtain them, I think with `getcontent` or `getcontents`. Alternatively, I think you could replace the for loop with a `map` that returns the axes (without decorations).

rikhuijzer: > I know, I think it would be more natural to do this in the loop since we only plot the figure after hiding the decorations but are not interested in any intermediate figures.

I don't understand this point. We're also not showing things at that in between point.

> You can obtain them, I think with getcontent or getcontents. Alternatively, I think you could replace the for loop with a map that returns the axes (without decorations).

I've considered the map and do syntax and know that you're a fan of it, but I find that it's hard to read for newish Julia users. That's why I'm using a list comprehension and a for loop. Both are much more familiar to most readers.

rikhuijzer: Can we agree to disagree on this one and merge as is?

rikhuijzer: Wait yes only the loop might be the easiest to read. I'm changing it now

rikhuijzer:
Ready for review again

cpfiffer (approved):
LGTM!

docs/src/makie.md
```diff
@@ -0,0 +1,76 @@
...
+end
+
+axes = [only(contents(fig[i, 2])) for i in 1:length(params)]
+linkxaxes!(axes...)
+
+fig
+```
+
+Finally, let's add a simple legend.
+Thanks to setting `label` above, this legend will have the right labels:
```
storopoli, line 70:
```suggestion
Due to the setting `label` above, this legend will have the right labels:
```

storopoli (approved):
Great work!

docs/src/makie.md
rikhuijzer: Cannot apply. There is a small typo in the sentence :stuck_out_tongue:

storopoli: Caffeine had not yet reached critical concentration in my bloodstream 🙄

PR #348, opened 2021-12-25
Update to MLJBase 0.19

devmotion:
Fixes #346. Fixes #347.


devmotion:
It seems MLJBase 0.19 breaks something: https://github.com/TuringLang/MCMCChains.jl/runs/4629683335?check_suite_focus=true#step:6:1201

devmotion:
Should be fixed by the different dispatches in https://github.com/TuringLang/MCMCDiagnosticTools.jl/pull/31:  MLJBase 0.19 uses probabilistic predictions that are not subtypes of `Distributions.Distribution`.

devmotion:
Tests pass now.

cpfiffer (approved):

PR #349, opened 2021-12-25
Make docstrings and implementation more consistent

rikhuijzer:


devmotion:
What's the motivation for this addition? Isn't it simpler to just index into the chain? You can just use `chain[:, param, chainindex]`. In contrast to `get` this should also be type stable (generally, type inference wise the implementation of `get` is quite bad since the type of the named tuple can't be informed from the vector of symbols).

rikhuijzer:
> What's the motivation for this addition? Isn't it simpler to just index into the chain? You can just use `chain[:, param, chainindex]`. In contrast to `get` this should also be type stable (generally, type inference wise the implementation of `get` is quite bad since the type of the named tuple can't be informed from the vector of symbols).

I didn't see that that was also possible. I've changed this PR to make docstrings and implementation more consistent.

src/chains.jl
```diff
@@ -219,8 +219,7 @@ julia> get(chn, :param_1; flatten=true)
 (param_1 = 1,)
 ```
 """
-Base.get(c::Chains, v::Symbol; flatten = false) = get(c, [v], flatten=flatten)
```
rikhuijzer, line 222: Moved this line down because it depends on the method below it. Makes reading easier.

cpfiffer (approved):
Seems perfectly fine to me -- nothing too big in here,

PR #350, opened 2021-12-26
Set version to 5.0.3

cpfiffer:


rikhuijzer (approved):

rikhuijzer:
I changed the name because more informative is more better in this case

I see that the MLJBase update (#348) is not registered yet, so I'm gonna go ahead and merge this PR and register a new version.

devmotion:
> I see that the MLJBase update (#348) is not registered yet, so I'm gonna go ahead and merge this PR and register a new version.

The update only fixes tests but doesn't change anything in the package. Only MCMCDiagnosticTools had to be fixed.

rikhuijzer:
> > I see that the MLJBase update (#348) is not registered yet, so I'm gonna go ahead and merge this PR and register a new version.
> 
> The update only fixes tests but doesn't change anything in the package. Only MCMCDiagnosticTools had to be fixed.

Also cleaning up confusion for readers about

image

rikhuijzer:
> The update only fixes tests but doesn't change anything in the package. Only MCMCDiagnosticTools had to be fixed.

But, you're right. I didn't notice that MLJBase was only updated in `test/Project.toml`. Registering this package was not necessary. I'll check better next time.

Issue #351, opened 2022-01-06
Why `naive_se`?


ParadaCarleton:
It seems a bit unusual to display `naive_se` by default, given it's strictly inferior to the MCSE measure. Mostly I'm concerned about people misunderstanding it. There's lots of things that `naive_se` could plausibly be confused with -- the standard error of the sample mean for the data, the standard deviation of the posterior distribution, or it might just be taken to be a good estimate of the MCSE, which it often isn't.

Another problem with `naive_se` came up in the course of adding support for weighted means, which are required for nested sampling. It's not at all clear to me what the "Naive" error here would be, since there's no autocorrelation.

Would it make sense to remove it?

PR #354, opened 2022-02-22
Weighted chains

ParadaCarleton:
A proposal to add a `weights` field to MCMCChains, for use with SMC, Importance Sampling, and Nested Samplers.


src/chains.jl
```diff
@@ -26,15 +26,16 @@ end
 
 # Generic chain constructor.
 function Chains(
-    val::AbstractArray{<:Union{Missing, Real},3},
+    val::AbstractArray{T,3},
     parameter_names::AbstractVector{Symbol} = Symbol.(:param_, 1:size(val, 2)),
-    name_map = (parameters = parameter_names,);
+    name_map = (parameters = parameter_names,),
+    weights=UnitWeights{T}(size(val, 1));
```
devmotion, line 32:
```suggestion
    weights=UnitWeights{Bool}(size(val, 1));
```

src/chains.jl
```diff
@@ -26,15 +26,16 @@ end
...
     start::Int = 1,
     thin::Int = 1,
     iterations::AbstractVector{Int} = range(start; step=thin, length=size(val, 1)),
     evidence = missing,
     info::NamedTuple = NamedTuple()
-)
+) where T<:Union{Missing, Real}
```
devmotion, line 38:
```suggestion
)
```

src/chains.jl
```diff
@@ -757,10 +756,24 @@ function _cat(::Val{1}, c1::Chains, args::Chains...)
...
+    weights = _weights_cat(c1.weights, getproperty.(args, :weights)...)
+
+    return Chains(value, missing, c1.name_map, c1.info, weights)
+end
 
-    return Chains(value, missing, c1.name_map, c1.info)
+function _weights_cat(w::UnitWeights...) 
+    # TODO: deal with case that weights are not one-dimensional
+    T = promote_type(eltype.(w)...)
+    return UnitWeights{T}(sum(x->size(x, 1), w))
```
devmotion, line 767: Avoid splatting and unnecessary allocations whenever possible:
```suggestion
    T = mapreduce(eltype, promote_type, w)
    return UnitWeights{T}(sum(length, w))
```

src/chains.jl
```diff
@@ -757,10 +756,24 @@ function _cat(::Val{1}, c1::Chains, args::Chains...)
...
 end
 
+function _weights_cat(w::UnitWeights...)
```
devmotion, line 770: I guess this is supposed to be
```suggestion
function _weights_cat(w::ProbabilityWeights...) 
```
?
BTW the current definitions of `_weights_cat` are problematic since they also define, and redefine, `_weights_cat()`.

src/chains.jl
```diff
@@ -757,10 +756,24 @@ function _cat(::Val{1}, c1::Chains, args::Chains...)
...
                       var = nms,
                       chain = chns)
+    weights = _weights_cat(c1.weights, getproperty.(args, :weights)...)
+
+    return Chains(value, missing, c1.name_map, c1.info, weights)
+end
 
-    return Chains(value, missing, c1.name_map, c1.info)
+function _weights_cat(w::UnitWeights...) 
+    # TODO: deal with case that weights are not one-dimensional
```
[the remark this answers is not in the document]
devmotion:
```suggestion
```

src/chains.jl
```diff
@@ -757,10 +756,24 @@ function _cat(::Val{1}, c1::Chains, args::Chains...)
                       iter = mapreduce(range, vcat, args; init=range(c1)),
                       var = nms,
                       chain = chns)
+    weights = _weights_cat(c1.weights, getproperty.(args, :weights)...)
```
devmotion, line 759: If possible, avoid splatting but use something similar to `mapreduce`. Maybe use
```suggestion
    weights = _weights_cat(c1, args)
```
and handle the weight extraction in `_weights_cat` (you can still dispatch on weight types by dispatching on the type parameter).

src/chains.jl
```diff
@@ -757,10 +756,24 @@ function _cat(::Val{1}, c1::Chains, args::Chains...)
...
+    T = promote_type(eltype.(w)...)
+    return UnitWeights{T}(sum(x->size(x, 1), w))
 end
 
+function _weights_cat(w::UnitWeights...) 
+    # TODO: deal with case that weights are not one-dimensional
+    w = reduce(vcat, w)
+    return ProbabilityWeights(w)
+end
+
```
devmotion, line 775: We should add a default `_weights_cat` method for `Weights`, and if supported, `AbstractVector`s.

devmotion: Maybe also good to provide a descriptive error message for incompatible types.

src/chains.jl
```diff
@@ -819,7 +837,8 @@ function _cat(::Val{3}, c1::Chains, args::Chains...)
     new_info = NamedTuple{tuple(nontime_props...)}(tuple([c1.info[n] for n in nontime_props]...))
     new_info = merge(new_info, (start_time = starts, stop_time = stops))
 
-    return Chains(value, missing, c1.name_map, new_info)
+    weights = UnitWeights{eltype(value)}(size(value, 1))
```
devmotion, line 840: The type should not be based on `value` (again, problems with `Missing`):
```suggestion
    weights = UnitWeights{Bool}(size(value, 1))
```

src/chains.jl
```diff
@@ -783,10 +796,15 @@ function _cat(::Val{2}, c1::Chains, args::Chains...)
...
-    return Chains(value, missing, name_map, c1.info)
+    return Chains(value, missing, name_map, c1.info, c1.weights)
 end
 
 function _cat(::Val{3}, c1::Chains, args::Chains...)
+    # TODO: Deal with the case that some chains have non-unit weights.
+    # Requires implementing array-like weights for StatsBase (#748)
+    if !all(c -> typeof(c.weights)<:UnitWeights, args)
+        throw(ArgumentError("all chains must have unit weights to concatenate"))
+    end
```
devmotion, line 807: This seems too restrictive? I guess it is sufficient if they are equal.
```suggestion
    # check weights
    w = c1.weights
    all(c -> c.weights == w, args) || throw(ArgumentError("all chains must have unit weights to concatenate"))
```

src/chains.jl
```diff
@@ -819,7 +837,8 @@ function _cat(::Val{3}, c1::Chains, args::Chains...)
     new_info = NamedTuple{tuple(nontime_props...)}(tuple([c1.info[n] for n in nontime_props]...))
     new_info = merge(new_info, (start_time = starts, stop_time = stops))
 
-    return Chains(value, missing, c1.name_map, new_info)
+    weights = UnitWeights{eltype(value)}(size(value, 1))
+    return Chains(value, missing, c1.name_map, new_info, weights)
```
devmotion, line 841: With the suggestion above, you can just write
```suggestion
    return Chains(value, missing, c1.name_map, new_info, w)
```

src/stats.jl
```diff
@@ -191,30 +191,39 @@ function hpd(chn::Chains; alpha::Real=0.05, kwargs...)
 end
 
 """
-    quantile(chains[; q = [0.025, 0.25, 0.5, 0.75, 0.975], append_chains = true, kwargs...])
+    quantile(chains, q = [0.025, 0.25, 0.5, 0.75, 0.975]; alpha, beta, kwargs...)
+    quantile(chains; q, alpha, beta, kwargs...)
```
devmotion, line 195: These changes are not required for this PR, are they?
```suggestion
    quantile(chains::Chains; q::AbstractVector = [0.025, 0.25, 0.5, 0.75, 0.975], kwargs...)
```

src/stats.jl
```diff
@@ -191,30 +191,39 @@ function hpd(chn::Chains; alpha::Real=0.05, kwargs...)
...
 
 Compute the quantiles for each parameter in the chain.
```
devmotion, line 197:
```suggestion
Compute the quantiles `q` for each parameter in the `chains`.
```

src/stats.jl
```diff
@@ -191,30 +191,39 @@ function hpd(chn::Chains; alpha::Real=0.05, kwargs...)
...
 
 Setting `append_chains=false` will return a vector of dataframes containing the quantiles
 for each chain.
+
+# Arguments
+
+- `chains`: A `Chains` object.
+- `q`: A vector of quantiles.
+- `alpha`, `beta`: See `?quantile` for details.
+- `kwargs...`: A set of keyword arguments to be passed to `summarize`.
```
devmotion, line 207:
```suggestion
```

src/stats.jl
```diff
@@ -191,30 +191,39 @@ function hpd(chn::Chains; alpha::Real=0.05, kwargs...)
...
 
 """
-    quantile(chains[; q = [0.025, 0.25, 0.5, 0.75, 0.975], append_chains = true, kwargs...])
+    quantile(chains, q = [0.025, 0.25, 0.5, 0.75, 0.975]; alpha, beta, kwargs...)
+    quantile(chains; q, alpha, beta, kwargs...)
 
 Compute the quantiles for each parameter in the chain.
 
 Setting `append_chains=false` will return a vector of dataframes containing the quantiles
 for each chain.
```
devmotion, line 200:
```suggestion
The remaining keyword arguments are forwarded to [`summarize`](@ref).
Setting `append_chains=false` will return a vector of dataframes containing the quantiles for each chain.
```

src/stats.jl
```diff
@@ -191,30 +191,39 @@ function hpd(chn::Chains; alpha::Real=0.05, kwargs...)
...
+- `alpha`, `beta`: See `?quantile` for details.
+- `kwargs...`: A set of keyword arguments to be passed to `summarize`.
 """
 function quantile(
     chains::Chains;
-    q::AbstractVector = [0.025, 0.25, 0.5, 0.75, 0.975],
-    append_chains = true,
+    q::AbstractVector=[0.025, 0.25, 0.5, 0.75, 0.975],
+    alpha::Real=0,
+    beta::Real=alpha,
```
devmotion, line 213:
```suggestion
```

src/stats.jl
```diff
@@ -191,30 +191,39 @@ function hpd(chn::Chains; alpha::Real=0.05, kwargs...)
...
     kwargs...
 )
     # compute quantiles
     funs = Function[]
     func_names = @. Symbol(100 * q, :%)
-    for i in q
-        push!(funs, x -> quantile(cskip(x), i))
+
+    for prob in q
+        push!(funs, x -> quantile(cskip(x), chains.weights, prob; alpha=alpha, beta=beta))
```
devmotion, line 221: This is not correct: `cskip` skips missing entries and hence the length of `cskip(x)` and `chains.weights` might not match. `alpha` and `beta` could be moved to a separate PR.

src/stats.jl
```diff
@@ -250,14 +259,19 @@ function summarystats(
     kwargs...
 )
     # Store everything.
-    funs = [mean∘cskip, std∘cskip, sem∘cskip, x -> MCMCDiagnosticTools.mcse(cskip(x); method=etype, kwargs...)]
-    func_names = [:mean, :std, :naive_se, :mcse]
+    funs = [
+        x -> mean(cskip(x), chains.weights), 
+        x -> std(cskip(x), chains.weights),
```
devmotion, line 264: Again these don't match in general.

src/stats.jl
```diff
@@ -250,14 +259,19 @@ function summarystats(
...
+    ]
+    func_names = [:mean, :std, :mcse]
 
     # Subset the chain.
     _chains = Chains(chains, _clean_sections(chains, sections))
 
     # Calculate ESS separately.
-    ess_df = MCMCDiagnosticTools.ess_rhat(_chains; sections = nothing, method = method, maxlag = maxlag)
+    # TODO: Fix ESS for weighted chains.
+    ess_df = MCMCDiagnosticTools.ess_rhat(_chains; sections=nothing, method=method, maxlag=maxlag)
```
devmotion, line 274: Unrelated change:
```suggestion
    ess_df = MCMCDiagnosticTools.ess_rhat(_chains; sections = nothing, method = method, maxlag = maxlag)
```

src/summarize.jl
```diff
@@ -134,13 +134,13 @@ Summarize `chains` in a `ChainsDataFrame`.
 * `summarize(chns; sections=[:parameters, :internals])` : Chain summary for multiple sections
 """
 function summarize(
-    chains::Chains, funs...;
+    chains::Chains{T,A,L,K,I,W}, funs::Base.Callable...;
```
devmotion, line 137: We don't want to restrict the function to `Base.Callable`. Probably it would also be better to dispatch on the weights in the function body when needed as large parts are the same, regardless of the weights. Maybe not even necessary to dispatch on the weights at all.
```suggestion
    chains::Chains{T,A,L,K,I,W}, funs...;
```

devmotion (changes requested):
Thanks for the PR, supporting weights properly is long overdue. I added some comments to the draft.

Project.toml
```diff
@@ -9,8 +9,10 @@ version = "5.0.3"
 AbstractMCMC = "80f14c24-f653-4e6a-9b94-39d6b0f70001"
 AxisArrays = "39de3d68-74b9-583c-8d2d-e117c070f3a9"
 Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
+ConcreteStructs = "2569d6c7-a4a2-43d3-a901-331e8e4be471"
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
 Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
+FillArrays = "1a297f60-69ca-5386-bcde-b61e274b549b"
```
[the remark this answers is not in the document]
ParadaCarleton: Yep, thanks -- I used those in early versions before removing them and forgot to remove from the reqs.

src/MCMCChains.jl
```diff
@@ -60,13 +61,22 @@ Parameters:
 - `info` : A `NamedTuple` containing miscellaneous information relevant to the chain.
 The `info` field can be set using `setinfo(c::Chains, n::NamedTuple)`.
 """
-struct Chains{T,A<:AxisArray{T,3},L,K<:NamedTuple,I<:NamedTuple} <: AbstractMCMC.AbstractChains
+struct Chains{T,A<:AxisArray{T,3},L,K<:NamedTuple,I<:NamedTuple,W<:AbstractWeights} <: AbstractMCMC.AbstractChains
```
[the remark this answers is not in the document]
ParadaCarleton: Will do.

src/chains.jl
```diff
@@ -671,9 +672,7 @@ Return a new `Chains` object with a `NamedTuple` type `n` placed in the `info` f
 new_chn = setinfo(chn, NamedTuple{(:a, :b)}((1, 2)))
 ```
 """
-function setinfo(c::Chains, n::NamedTuple)
-    return Chains(c.value, c.logevidence, c.name_map, n)
-end
+setinfo(c::Chains, n::NamedTuple) = Chains(c.value, c.logevidence, c.name_map, n, c.weights)
```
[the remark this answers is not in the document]
ParadaCarleton: Makes sense.

src/chains.jl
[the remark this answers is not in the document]
ParadaCarleton: Currently that's correct, but in the future I'd like to deal with this by making a PR to StatsBase allowing for [general array-shaped weights](https://github.com/JuliaStats/StatsBase.jl/issues/748). That way we can have multiple chains where each chain has its own associated weights.

src/chains.jl
ParadaCarleton: Is there any situation where we want to concatenate chains with non-unit weights?

src/chains.jl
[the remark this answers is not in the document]
devmotion: Yes, I saw the issue. I think it is better to open an issue in MCMCChains as a reminder, I don't think people regularly check the source code for TODOs :wink:

src/chains.jl
devmotion: I don't know - but not even unit weights have to be represented as `UnitWeights`, e.g., it could also be of type `Weights` with an array of equal weights. Not checking the types but the values (which uses shortcuts for special types such as `UnitWeights`: https://github.com/JuliaStats/StatsBase.jl/blob/071d10a110d756fa15c6a6878936eee27e4c2cb6/src/weights.jl#L369) works much more generally and also with weight types not defined in StatsBase.

src/chains.jl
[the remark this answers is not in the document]
ParadaCarleton: I can open an issue as well -- the TODOs are mostly there to make it easier to find where in the code I need to make changes.

src/chains.jl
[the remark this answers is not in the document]
ParadaCarleton: Actually, I think this does raise the question of why `Chains` are immutable. I'm not sure there's many practical situations where we'll be creating enough `Chains` to mess with the garbage collecter, and there's nothing you can really stack-allocate anyways because everything is stuffed in an array. Is there something I'm missing?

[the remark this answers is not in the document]
devmotion: What would be the advantage of them being mutable? It seems it would make it easier to violate the constraints that we impose on the different fields. E.g., there are assumptions on the `range` of a `Chains` which could be broken easily. Since we don't copy but share the underlying data immutable `Chains` shouldn't cause (many) more allocations than a mutable alternative.

Project.toml
```diff
@@ -3,10 +3,11 @@ uuid = "c7f686f2-ff18-58e9-bc7b-31028e88f75d"
 keywords = ["markov chain monte carlo", "probablistic programming"]
 license = "MIT"
 desc = "Chain types and utility functions for MCMC simulations."
-version = "5.1.0"
+version = "6.0.0"
 
 [deps]
 AbstractMCMC = "80f14c24-f653-4e6a-9b94-39d6b0f70001"
+ArraysOfArrays = "65a8f2f4-9b39-5baf-92e2-a9cc46fdf018"
```
devmotion, line 10: Why is this needed? Weights should not be related to ArraysOfArrays (which BTW is only one possible approach to arrays of arrays).

src/chains.jl
```diff
@@ -26,9 +26,10 @@ end
 
 # Generic chain constructor.
 function Chains(
-    val::AbstractArray{<:Union{Missing, Real},3},
+    val::AbstractArray{<:Union{Missing,Real},3},
     parameter_names::AbstractVector{Symbol} = Symbol.(:param_, 1:size(val, 2)),
     name_map = (parameters = parameter_names,);
+    weights=StatsBase.UnitWeights{Bool}(size(val, 1)),  # TODO: Multiple weighted chains
```
devmotion, line 32: This comment is quite confusing IMO. I suggest removing it and opening a more concrete issue instead since nobody will browse the code for TODOs:
```suggestion
    weights=StatsBase.UnitWeights{Bool}(size(val, 1)),
```
Alternatively, clarify it and write e.g.
```suggestion
    weights=StatsBase.UnitWeights{Bool}(size(val, 1)),  # TODO: Different weights per chain (StatsBase only supports weight vectors currently)

File: src/chains.jl [hunk 25]
```diff
@@ -76,9 +77,21 @@ function Chains(
...
+    logevidence::Union{Missing,Real}, 
+    name_map::NamedTuple, 
+    info::NamedTuple
+)
+    return Chains(
+        value, logevidence, name_map, info, StatsBase.UnitWeights{Bool}(size(value, 1))
+    )
+end
+
+
```
devmotion, line 94: I don't think this is needed. More constructors without clear benefit mainly creates confusion about what is the "official" way to construct it. In general, IMO keyword arguments are clearer.
```suggestion
```

src/chains.jl
devmotion: Can you address this comment and avoid splatting?

src/chains.jl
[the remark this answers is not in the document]
devmotion: I don't think there's a clear benefit from having these TODOs - you basically have to go through and possibly change the few methods that deal with weights for such a change.
```suggestion
```

src/chains.jl
```diff
@@ -757,10 +756,24 @@ function _cat(::Val{1}, c1::Chains, args::Chains...)
...
 
-    return Chains(value, missing, c1.name_map, c1.info)
+function _weights_cat(w::UnitWeights...) 
+    # TODO: deal with case that weights are not one-dimensional
+    T = promote_type(eltype.(w)...)
+    return UnitWeights{T}(sum(x->size(x, 1), w))
 end
 
+function _weights_cat(w::UnitWeights...) 
+    # TODO: deal with case that weights are not one-dimensional
```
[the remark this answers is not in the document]
devmotion:
```suggestion
```

src/chains.jl
devmotion: Can you fix this issue?

src/stats.jl
```diff
@@ -191,30 +191,29 @@ function hpd(chn::Chains; alpha::Real=0.05, kwargs...)
...
-for each chain.
+The remaining keyword arguments are forwarded to [`summarize`](@ref).
+Setting `append_chains=false` will return a vector of dataframes containing the quantiles for each chain.
 """
-function quantile(
+function StatsBase.quantile(
     chains::Chains;
-    q::AbstractVector = [0.025, 0.25, 0.5, 0.75, 0.975],
-    append_chains = true,
+    q::AbstractVector=[0.025, 0.25, 0.5, 0.75, 0.975],
```
devmotion, line 203: Unrelated change:
```suggestion
    q::AbstractVector = [0.025, 0.25, 0.5, 0.75, 0.975],
```

src/stats.jl
```diff
@@ -191,30 +191,29 @@ function hpd(chn::Chains; alpha::Real=0.05, kwargs...)
...
     kwargs...
```
devmotion, line 204: Is `append_chains=true` still the default?

src/stats.jl
```diff
@@ -191,30 +191,29 @@ function hpd(chn::Chains; alpha::Real=0.05, kwargs...)
...
 )
     # compute quantiles
     funs = Function[]
     func_names = @. Symbol(100 * q, :%)
-    for i in q
-        push!(funs, x -> quantile(cskip(x), i))
+
+    for prob in q
+        push!(funs, Base.Fix2(StatsBase.quantile, prob))
```
devmotion, line 211: This does not skip missing values anymore, it seems. Can you fix this?

src/stats.jl
```diff
@@ -228,8 +227,6 @@ end
         append_chains= true,
         method::AbstractESSMethod = ESSMethod(),
         maxlag = 250,
-        etype = :bm,
-        kwargs...
```
devmotion, line 232: It seems `kwargs...` are still supported and should not be removed? Is `etype=:bm` still the default?

src/stats.jl
```diff
@@ -246,18 +243,24 @@ function summarystats(
     append_chains::Bool = true,
     method::MCMCDiagnosticTools.AbstractESSMethod = ESSMethod(),
     maxlag = 250,
-    etype = :bm,
+    skip_missing=true,
```
devmotion, line 246: This is not used?

src/stats.jl
```diff
@@ -246,18 +243,24 @@ function summarystats(
...
 )
     # Store everything.
-    funs = [mean∘cskip, std∘cskip, sem∘cskip, x -> MCMCDiagnosticTools.mcse(cskip(x); method=etype, kwargs...)]
-    func_names = [:mean, :std, :naive_se, :mcse]
+    funs = [
+        mean, 
+        std,
+        MCMCDiagnosticTools.mcse,
+    ]
+    func_names = [:mean, :std, :mcse]
```
devmotion, line 255: These changes seem unrelated and undesired. Can you revert them?

src/stats.jl
```diff
@@ -246,18 +243,24 @@ function summarystats(
...
 
     # Subset the chain.
     _chains = Chains(chains, _clean_sections(chains, sections))
 
     # Calculate ESS separately.
-    ess_df = MCMCDiagnosticTools.ess_rhat(_chains; sections = nothing, method = method, maxlag = maxlag)
+    ess_df = MCMCDiagnosticTools.ess_rhat(
+        _chains; sections = nothing, method = method, maxlag = maxlag
+    )
```
devmotion, line 263: Unrelated changes:
```suggestion
    ess_df = MCMCDiagnosticTools.ess_rhat(_chains; sections = nothing, method = method, maxlag = maxlag)
```

src/stats.jl
```diff
@@ -279,7 +282,7 @@ Calculate the mean of a chain.
 """
 function mean(chains::Chains; kwargs...)
     # Store everything.
-    funs = [mean∘cskip]
+    funs = [mean]
```
devmotion, line 285: Unrelated:
```suggestion
    funs = [mean∘cskip]
```

src/summarize.jl
```diff
@@ -134,16 +134,18 @@ Summarize `chains` in a `ChainsDataFrame`.
 * `summarize(chns; sections=[:parameters, :internals])` : Chain summary for multiple sections
 """
 function summarize(
-    chains::Chains, funs...;
+    chains::Chains, funs::Base.Callable...;
```
devmotion, line 137:
```suggestion
    chains::Chains, funs...;
```

src/summarize.jl
```diff
@@ -134,16 +134,18 @@ Summarize `chains` in a `ChainsDataFrame`.
...
+    skip_missing=true,
```
devmotion, line 138:
```suggestion
```

src/summarize.jl
```diff
@@ -134,16 +134,18 @@ Summarize `chains` in a `ChainsDataFrame`.
...
     sections = _default_sections(chains),
-    func_names::AbstractVector{Symbol} = Symbol[],
-    append_chains::Bool = true,
-    name::String = "",
+    func_names::AbstractVector{Symbol}=Symbol[],
```
devmotion, line 140:
```suggestion
    func_names::AbstractVector{Symbol} = Symbol[],
```

src/summarize.jl
```diff
@@ -134,16 +134,18 @@ Summarize `chains` in a `ChainsDataFrame`.
...
+    append_chains::Bool=true,
```
devmotion, line 141:
```suggestion
    append_chains::Bool = true,
```

src/summarize.jl
```diff
@@ -134,16 +134,18 @@ Summarize `chains` in a `ChainsDataFrame`.
...
+    name::String="",
```
devmotion, line 142:
```suggestion
    name::String = "",
```

src/summarize.jl
```diff
@@ -134,16 +134,18 @@ Summarize `chains` in a `ChainsDataFrame`.
...
     additional_df = nothing
 )
+
```
devmotion, line 145:
```suggestion
```

src/summarize.jl
```diff
@@ -134,16 +134,18 @@ Summarize `chains` in a `ChainsDataFrame`.
...
     # If we weren't given any functions, fall back to summary stats.
     if isempty(funs)
-        return summarystats(chains; sections = sections)
+        return summarystats(chains; sections=sections)
```
devmotion, line 148:
```suggestion
        return summarystats(chains; sections = sections)
```

src/summarize.jl
```diff
@@ -158,10 +160,16 @@ function summarize(
     # Obtain the additional named tuple.
     additional_nt = additional_df === nothing ? NamedTuple() : additional_df.nt
 
+    # TODO: Handle multiple weighted chains.
```
devmotion, line 163:
```suggestion
```

src/summarize.jl
```diff
@@ -158,10 +160,16 @@ function summarize(
...
     if append_chains
         # Evaluate the functions.
         data = to_matrix(chn)
-        fvals = [[f(data[:, i]) for i in axes(data, 2)] for f in funs]
+
+        if chains.weights isa StatsBase.UnitWeights
+            fvals = [[f(data[:, i]) for i in axes(data, 2)] for f in funs]
+        else 
+            fvals = [[f(data[:, i], chains.weights) for i in axes(data, 2)] for f in funs]
+        end
```
devmotion, line 172: It would be easier here to handle this by dispatching on `weights`.
```suggestion
        fvals = [[f(data[:, i], chains.weights) for i in axes(data, 2)] for f in funs]
```

src/summarize.jl
```diff
@@ -171,8 +179,18 @@ function summarize(
...
+            vector_of_fvals = [
+                [[f(x[:, i]) for i in axes(x, 2)] for f in funs] for x in data
+            ]
+        else
+            vector_of_fvals = [
+                [[f(x[:, i], chains.weights) for i in axes(x, 2)] for f in funs] 
+                for x in data
+            ]
+        end
+
```
devmotion, line 193:
```suggestion
        vector_of_fvals = [[[f(x[:, i], chains.weights) for i in axes(x, 2)] for f in funs] for x in data]

```

test/summarize_tests.jl
```diff
@@ -18,16 +18,16 @@ using Statistics: std
     show(stdout, "text/plain", parm_df)
 
     @test 0.48 < parm_df[:a, :mean][1] < 0.52
-    @test names(parm_df) == [:parameters, :mean, :std, :naive_se, :mcse, :ess, :rhat]
+    @test names(parm_df) == [:parameters, :mean, :std, :mcse, :ess, :rhat]
```
devmotion, line 21:
```suggestion
    @test names(parm_df) == [:parameters, :mean, :std, :naive_se, :mcse, :ess, :rhat]
```

test/summarize_tests.jl
```diff
@@ -18,16 +18,16 @@ using Statistics: std
...
 
     # Indexing tests
     @test convert(Array, parm_df[:a, :]) == convert(Array, parm_df[:a])
     @test parm_df[:a, :][:,:parameters] == :a
     @test parm_df[[:a, :b], :][:,:parameters] == [:a, :b]
 
     all_sections_df = summarize(chns, sections=[:parameters, :internals])
-    @test all_sections_df[:,:parameters] == [:a, :b, :c, :d, :e, :f, :g, :h]
-    @test size(all_sections_df) == (8, 7)
+    @test all_sections_df[:, :parameters] == [:a, :b, :c, :d, :e, :f, :g, :h]
```
devmotion, line 29:
```suggestion
    @test all_sections_df[:,:parameters] == [:a, :b, :c, :d, :e, :f, :g, :h]
```

test/summarize_tests.jl
```diff
@@ -18,16 +18,16 @@ using Statistics: std
...
+    @test size(all_sections_df) == (8, 6)
```
devmotion, line 30:
```suggestion
    @test size(all_sections_df) == (8, 7)
```

devmotion (changes requested):
Thanks for the PR.

Please avoid any unrelated changes and do not remove any existing functionality in this PR. Can you please also add tests that cover all new functions and all weights functionality?

Project.toml
ParadaCarleton: Accidental, thanks.

src/chains.jl
ParadaCarleton: This just defaults to unit weights, to avoid breaking any old code that creates `Chains` without including weights. If I don't include it, some of the tests break.

src/stats.jl
ParadaCarleton: This is related, `cskip` doesn't work with weights.

src/summarize.jl
ParadaCarleton: As in, the user should pass functions that ignore the passed weights? That would definitely be easier but causes problems for functions that don't have a method taking `weights`.

src/chains.jl
devmotion: The PR is breaking anyway. IMO it is much better to fix the tests than to add an additional constructor.

src/stats.jl
ParadaCarleton: This is related, `cskip` doesn't work with weights, and `mcse` needs to be passed like this to make sure it works either with or without weights.

`naive_se` was removed because there's not really a way to define the naive standard error for a weighted chain (since weighted chains generally consist of IID samples).

devmotion: > This is related, cskip doesn't work with weights, and mcse needs to be passed like this to make sure it works either with or without weights.

That's why I asked you to fix it 🙂 We absolutely want to keep the current behaviour since otherwise in realistic scenarios, such as e.g. the Dirichlet process example IIRC, summary statistics will just be missing.

> naive_se was removed because there's not really a way to define the naive standard error for a weighted chain (since weighted chains generally consist of IID samples). Is there an actual use case for the naive standard errors, given they're inconsistent?

We should discuss this issue separately but not include such changes in this PR. Ie we should keep it for the unweighted/uniformly weighted case.
a

src/stats.jl
devmotion: See my comment above.

src/summarize.jl
devmotion: All functions that we use in MCMCChains have to support weights anyway before this PR is merged. And I think it's reasonable to demand the same from any user-provided functions since one can just ignore the weights if one only uses unweighted samples (not sure if anyone actually uses this function and if it should be exported but that's a separate issue).

src/stats.jl
ParadaCarleton: > > This is related, cskip doesn't work with weights, and mcse needs to be passed like this to make sure it works either with or without weights.
> 
> That's why I asked you to fix it

Sorry if I was a bit unclear. What I meant is that _`cskip`_ doesn't work with weights, because `cskip` only takes one argument, not that it's impossible to skip missing weights. The most recent PR fixes this by skipping missing observations without using `cskip`.

ParadaCarleton: > We should discuss this issue separately but not include such changes in this PR. Ie we should keep it for the unweighted/uniformly weighted case.

I created another issue where this can be discussed, #351 , if you'd like. If you'd like I can create a different PR removing `naive_se` if you don't want to combine them.

src/summarize.jl
ParadaCarleton: I mean, maybe it shouldn't be, but ATM it is and I'd like to minimize the amount of code we break.

devmotion: It's also easier for downstream packages if we reduce the number of breaking releases.

PR #365, opened 2022-04-29
Fix sampling from multiple chains

### LINKED ISSUE #223: Unexpected number of samples returned from `sample`
The `sample` parameter `n` determines the number of samples to retrieve. However, it returns more than that for "multi" chains. Consider the following:

```julia
using Turing, MCMCChains

@model gdemo(x, y) = begin
    σ ~ InverseGamma(2,3)
    μ ~ Normal(0,sqrt(σ))

    x ~ Normal(μ, sqrt(σ))
    y ~ Normal(μ, sqrt(σ))
end
s = sample(gdemo(1, 2), HMC(0.01, 5), MCMCThreads(), 2, 4)

x = 4
(rows, _) = size(Array(sample(s, x)))
@assert rows == x
```

I was expecting it to return `n` number of samples per parameter, not `n * JULIA_NUM_THREADS`. Is this behaviour intended?

devmotion:
Fixes #223. This came up also in https://github.com/TuringLang/TuringTutorials/pull/296.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -3,7 +3,7 @@ uuid = "c7f686f2-ff18-58e9-bc7b-31028e88f75d"
 keywords = ["markov chain monte carlo", "probablistic programming"]
 license = "MIT"
 desc = "Chain types and utility functions for MCMC simulations."
-version = "5.1.1"
+version = "5.1.2"
 
 [deps]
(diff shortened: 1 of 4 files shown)


cpfiffer (approved):
LFTM!

PR #371, opened 2022-06-02
Don't warn if start or stop times are missing from a chain

chriselrod:
The warning seems unnecessarily intrusive.


src/chains.jl
```diff
@@ -583,9 +583,14 @@ function header(c::Chains; section=missing)
...
-    compute = compute_duration(c)
-
+    if Base.sym_in(:start_time, keys(c.info)) &&
+        Base.sym_in(:stop_time, keys(c.info))
+        wall = wall_duration(c)
+        compute = compute_duration(c)
+    else
+        wall = missing
+        compute = missing
+    end
```
devmotion, line 593: IMO it would be better to just remove the warnings in `to_datetime`. The functions already return the correct values, ie, `missing`, here which clearly indicate that the timings were missing. I don't see why they would ever be useful.

chriselrod: Okay. I figured maybe the intent of also printing the type was in case someone passed a `MyCustomButCurrentlyUnsuportedTimeType` object.

chriselrod: I've now instead removed the `@warn`.

devmotion: I'm not completely sure what @cpfiffer's (IIRC) original intention was. If we want to keep a warning for unknown types we could alternatively add a method for `::Missing` without warning which would have the same effect in the `show` method here. But since we also don't throw a warning in `to_datetime_vec` but just return `[missing]` it feels fine to remove the warning here as well, in my opinion.

src/chains.jl
```diff
@@ -375,7 +375,6 @@ max_datetime(ts) = maximum(to_datetime_vec(ts))
 
 # does not specialize on `typeof(T)`
 function missing_datetime(T::Type)
-    @warn "timestamp of type $(T) unknown"
     return missing
 end
```
devmotion, line 379: It seems this method is only called by `to_datetime` above? In that case, could we just remove it conpletely and directly return `missing` in `to_datetime`?

chriselrod: We'd have to pick a point to intercept. Currently, I now have `to_datetime_vec` return `missing`, and then `(min/max)_datetime` forward `missing`s.

This is because of calls like
```julia
min_start(c::Chains) = min_datetime(start_times(c))
```
`start_times` will return `missing`, and then `min_datetime` must handle it.

We could also change the definition of `min_start` to check whether `info` contains the appropriate field.

src/chains.jl
```diff
@@ -364,21 +364,16 @@ Base.convert(::Type{Array}, chn::Chains) = convert(Array, chn.value)
 # timestamps.
 to_datetime(t::DateTime) = t
 to_datetime(t::Float64) = unix2datetime(t)
-to_datetime(t) = missing_datetime(typeof(t))
 to_datetime_vec(t::Union{Float64, DateTime}) = [to_datetime(t)]
 to_datetime_vec(t::DateTime) = [to_datetime(t)]
 to_datetime_vec(ts::Vector) = map(to_datetime, ts)
-to_datetime_vec(ts) = [missing]
+to_datetime_vec(ts) = missing
```
devmotion, line 370: This change seems a bit dangerous. AFAICT the method is supposed to return a vector, as in the cases above and currently done. It should also not be necessary for getting rid of the warnings. What was your motivation for this change?


```suggestion
to_datetime_vec(ts) = [missing]
```

chriselrod: Those were the only calls `to_datetime_vec`, so the motivation was to avoid allocation `[missing]`.

PR #383, opened 2022-09-18
Compat 4

chriselrod:


devmotion:
Some dependency holds it back, IIRC do the packages in the JuliaFolds org not support Compat 4 yet. I'll close this PR and keep the CompatHelper PR open that we can monitor and rerun if the dependencies are updated.

chriselrod:
> 

Okay. I just made the PRs that I needed locally to be able to install packages like Pumas when using a custom system image on Julia 1.8 that had Compat 4.2 and DocStringExtensions.
BangBang and MetaTheory were the only two other packages that needed updates, so I made PRs for them as well.

devmotion:
Yeah, unfortunately it seems there are quite a few open CompatHelper PRs in JuliaFolds (e.g., also Transducers does not support Setfield 1 yet and hence we can't test and use it in DynamicPPL/Turing). I wrote a comment on Slack (https://julialang.slack.com/archives/C67910KEH/p1663534616775329?thread_ts=1663322677.862739&cid=C67910KEH), maybe that can speed up the required updates.

PR #392, opened 2022-12-13
Support MCMCDiagnosticTools v0.2

sethaxen:
This PR supersedes #391 in bumping compat for MCMCDiagnostocTools to v0.2. Since that breaking change changed the dimension order, we need to also drop compat for v0.1. v0.2's Julia lower bound is v1.3, so I've also set v1.3 as our minimum supported version.

The only other thing I did was add an undocumented `sections` keyword argument to `rstar`, similar to `ess_rhat`.


sethaxen:
Ugh, it looks like on v1.3 this is hitting dependency conflict hell.

devmotion:
I wonder if these dependency issues are not new but also present on the master branch. Maybe we just did not realize that there are problems since we only tested 1.0 and 1.8?

sethaxen:
It seems the dependency conflicts already existed with MCMCDiagnosticTools v0.1.5. So if we look at the past test CI, only v0.1.4 is being pulled in. v0.1.5 bumped the required version of MLJModelInterface from v1 to v1.6 in https://github.com/TuringLang/MCMCDiagnosticTools.jl/pull/44, which apparently causes the conflict we're now seeing.

devmotion:
Hmm annoying.

> Edit: but it's really not clear to me if there's an outdated compat entry somewhere in the ecosystem that could be fixed to make everything work.

It's not clear to me either. I'm not even sure what's exactly the problem. Is it
```
Unsatisfiable requirements detected for package StatsPlots [f3b207a7]:
StatsPlots [f3b207a7] log:
 ├─possible versions are: [0.10.0-0.10.2, 0.11.0, 0.12.0, 0.13.0, 0.14.0-0.14.34, 0.15.0-0.15.4] or uninstalled
 ├─restricted to versions 0.14.17-0.15 by an explicit requirement, leaving only versions [0.14.17-0.14.34, 0.15.0-0.15.4]
 ├─restricted by julia compatibility requirements to versions: [0.10.0-0.10.2, 0.11.0, 0.12.0, 0.13.0, 0.14.0-0.14.34] or uninstalled, leaving only versions: 0.14.17-0.14.34
 └─restricted by compatibility requirements with Interpolations [a98d9a8b] to versions: [0.10.0-0.10.2, 0.11.0, 0.15.1-0.15.4] or uninstalled — no versions left
   └─Interpolations [a98d9a8b] log:
     ├─possible versions are: [0.3.6-0.3.8, 0.4.0, 0.9.0-0.9.2, 0.10.1-0.10.6, 0.11.0-0.11.2, 0.12.0-0.12.10, 0.13.0-0.13.6, 0.14.0-0.14.7] or uninstalled
     └─restricted to versions 0.14.7 by an explicit requirement, leaving only versions 0.14.7
```
? In that case Interpolations seems problematic?

I'm not sure how much time we should spend on investigating this further. Maybe we should move to 1.6 instead?

devmotion:
Seems tests pass on Julia 1.6.

@cpfiffer are you fine with dropping support for Julia < 1.6? Or do you think we should try to investigate this strange dependency issue a bit more?

cpfiffer:
Oh, thanks for asking!

I'd be perfectly fine with dropping support for <1.6.

devmotion (approved):
Looks good to me :+1:

PR #398, opened 2023-01-30
Escape regex in namesingroup properly

torfjelde:
Adds proper escaping for indexing in `namesingroup`, hence adding support for stuff like `chain[:, Symbol("A[1]"), :]`.

### DIFF
--- a/src/chains.jl
+++ b/src/chains.jl
@@ -161,7 +161,7 @@ function namesingroup(chains::Chains, sym::Symbol; index_type::Symbol=:bracket)
     idx_str = index_type == :bracket ? "[" : "."
     # Start by looking up the symbols in the list of parameter names.
     names_of_params = names(chains)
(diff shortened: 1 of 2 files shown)


src/chains.jl
```diff
@@ -155,7 +161,7 @@ function namesingroup(chains::Chains, sym::Symbol; index_type::Symbol=:bracket)
     idx_str = index_type == :bracket ? "[" : "."
     # Start by looking up the symbols in the list of parameter names.
     names_of_params = names(chains)
-    regex = Regex("^$sym\$|^$sym\\$idx_str")
+    regex = Regex("^\\Q$sym\\E\$|^\\Q$sym\\E\\$idx_str")
```
devmotion, line 164: Wouldn't it be even better to use
```suggestion
    regex = Regex("^\\Q$sym\\E\$|^\\Q$sym$idx_str\\E")
```
?

PR #403, opened 2023-02-27
Revert "Fix MLJDecisionTreeInterface to 0.3.0"

devmotion:
Reverts TuringLang/MCMCChains.jl#402 now that the issue is fixed in MCMCDiagnosticTools.


devmotion:
Has to be included in #401, of course.

PR #412, opened 2023-04-16
Allow show to function properly even if some statistic computations fail

torfjelde:
In some scenarios where the chain is producing weird values, we can easily run into issues with some of the statistics we compute in `show`, which means that `show` simply breaks. This can be both very annoying and confusing. IMO `show` shouldn't break just because we have numerical issues for _one_ of the computations.

This PR adds a `try`/`catch` around the problematic computations, so that `show` can still work even if some of the statistics fail to compute.

### DIFF
--- a/src/stats.jl
+++ b/src/stats.jl
@@ -321,6 +321,7 @@ function summarystats(
     try
         ess_rhat_rank_df = MCMCDiagnosticTools.ess_rhat(
             _chains; sections = nothing, autocov_method = autocov_method, maxlag = maxlag, kind=:rank
+        )
         nt_ess_rhat_rank = (
             ess_bulk=ess_rhat_rank_df.nt.ess,
             rhat=ess_rhat_rank_df.nt.rhat,
--- a/src/stats.jl
+++ b/src/stats.jl
@@ -304,7 +304,7 @@ function summarystats(
         mcse_df = MCMCDiagnosticTools.mcse(
             _chains; sections = nothing, autocov_method = autocov_method, maxlag = maxlag,
         )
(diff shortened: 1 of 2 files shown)


cpfiffer (approved):

cpfiffer:
I like this -- will merge in after the tests pass.

src/stats.jl
```diff
@@ -299,23 +299,41 @@ function summarystats(
...
+            ess_per_sec=ess_rhat_rank_df.nt.ess_per_sec
+        )
+        nt_additional = merge(nt_additional, nt_ess_rhat_rank)
+    catch e
+        @warn "Bulk ESS/R-hat calculation failed: $e"
+    end
+
+    # Possibly re-order the columns to stay backwards-compatible.
+    additional_keys = (:mcse, :ess_bulk, :ess_tail, :rhat, :ess_per_sec)
+    additional_df = ChainDataFrame("Additional", (; ((k, nt_additional[k]) for k in additional_keys if k ∈ keys(nt_additional))...))
```
cpfiffer, line 336: The tests are failing due to a mis-ordering, I think all we have to do is this

```suggestion
    # Possibly re-order the columns to stay backwards-compatible.
    additional_keys = (:mcse, :ess_tail, :ess_bulk, :rhat, :ess_per_sec)
    additional_df = ChainDataFrame("Additional", (; ((k, nt_additional[k]) for k in additional_keys if k ∈ keys(nt_additional))...))
```

torfjelde: Nah, I think they're correct now:) I just added this piece of code; when you approved it I hadn't added this line.

Issue #414, opened 2023-04-16
`corner` fails with certain backends


torfjelde:
`corner` currently passes the parameters as-is, i.e. they will be `Symbol` rather than `string`. This can cause issues with certain backends as it's technically a violation of what Plots.jl expects (AFAIK).

E.g. `unicodeplots` will break when using `corner`.

torfjelde:
Solved by #415

PR #415, opened 2023-04-16
Fixes for `corner`

### LINKED ISSUE #413: `corner` breaks because of `chains` usage
Currently, something like

``` julia
corner(chain[:, :, 2:3])
```

breaks because we use `chains(chain)` in the `corner` recipe:

https://github.com/TuringLang/MCMCChains.jl/blob/ddac60fb57b8dba6955922cec46573b310dbc0c1/src/plot.jl#L187

This, IMO, also points to a slightly larger issue: `chains` can potentially return a sequence of indices that we cannot use to index into the given chain. For example, `chain` :shrug:

Not entirely certain if we should change how `chain` works or not.

torfjelde:
Fixes #413 and #414.


cpfiffer (approved):

cpfiffer:
Thanks Tor!

PR #423, opened 2023-05-26
Version bump

yebai:


devmotion (commented):
Yes, we should make a new release but I think it should include #422.

devmotion (approved):
I included the fix of the test compat entry. Looks good to me if tests pass.

PR #428, opened 2023-07-11
[Feature] Probability of Direction (pd)

DominiqueMakowski:
Indices of effect significance/existence computed from posterior distributions are very common in some fields. This feature adds the [probability of direction](https://en.wikipedia.org/wiki/Probability_of_direction), `p_direction()`, a Bayesian [equivalent](https://easystats.github.io/bayestestR/articles/probability_of_direction.html#relationship-with-the-p-value) of the p-value. 

I followed the implementation of `mean()` by using the `summarize()` workflow.

### DIFF
--- a/src/significance.jl
+++ b/src/significance.jl
@@ -1,0 +1,19 @@
+function p_direction(x::Vector{Float64})
+    return maximum([sum(x .> 0) ./ length(x), sum(x .< 0) ./ length(x)])
+end
+
+function p_direction(chains::Chains; kwargs...)
+    # Store everything.
+    funs = [p_direction]
+    func_names = [:p_direction]
+
+    # Summarize.
+    summary_df = summarize(
+        chains, funs...;
+        func_names=func_names,
+        name="Probability of Direction (pd)",
+        kwargs...
+    )
+
+    return summary_df
+end
--- a/src/MCMCChains.jl
+++ b/src/MCMCChains.jl
@@ -48,6 +48,7 @@ export rafterydiag
 export rstar
 
 export hpd
+export p_direction
 
 """
     Chains
@@ -85,5 +86,6 @@ include("modelstats.jl")
 include("plot.jl")
 include("tables.jl")
 include("rstar.jl")
+include("significance.jl")
 
 end # module


devmotion:
Thank you for the PR! I'd suggest, however, opening a PR to MCMCDiagnosticTools (first) - most (all?) diagnostics specialized in MCMCChains for `Chains` objects are actually owned by MCMCDiagnosticTools and defined there for chains in a generic `AbstractArray` format. The implementations in MCMCDiagnosticTools are not only used by MCMCChains but e.g. also by ArviZ.jl.

DominiqueMakowski:
Should I do that even though it's not technically a "diagnostic" index? It's more a summary index, similar to HPD.

If so, could you point me to an example of a function implemented there that I can follow?

devmotion:
I think MCMCDiagnosticTools would be preferable because it can be re-used easily by ArviZ, MCMCChains, and other array-based chain structures. The implementation would be even simpler than in this PR as you don't have to deal with the `Chains` internals, you would just return these values for a `::AbstractArray{<:Union{Missing,Real}}` chain(s) of shape `(draws[, chains[, params...]])` (see e.g. https://github.com/TuringLang/MCMCDiagnosticTools.jl/blob/main/src/mcse.jl).

src/significance.jl
```diff
@@ -0,0 +1,19 @@
+function p_direction(x::Vector{Float64})
+    return maximum([sum(x .> 0) ./ length(x), sum(x .< 0) ./ length(x)])
```
devmotion, line 2: Hmm, this creates a few unnecessary allocations (the array inside of `maximum`, `x .> 0`, and `x .< 0`). This could be fixed e.g. by
```suggestion
    ntotal = 0
    npositive = 0
    nnegative = 0
    for xi in x
        if xi > 0
            npositive += 1
        elseif xi < 0
            nnegative += 1
        end
        ntotal += 1
    end
    return max(npositive, nnegative) / ntotal
```

src/significance.jl
```diff
@@ -0,0 +1,19 @@
...
+
+function p_direction(chains::Chains; kwargs...)
+    # Store everything.
+    funs = [p_direction]
+    func_names = [:p_direction]
+
+    # Summarize.
+    summary_df = summarize(
+        chains, funs...;
+        func_names=func_names,
```
devmotion, line 13: To avoid alloctions:
```suggestion
    # Summarize.
    summary_df = summarize(
        chains, p_direction;
        func_names=[:p_direction],
```

devmotion (commented):
I added a few comments that I think would be good to address, regardless of where the functionality is added.

Moreover, this would need a few tests.

src/significance.jl
DominiqueMakowski: I am in disbelief that this code is more efficient than the above one-liner, but I checked with `@btime` and indeed... 🤯

devmotion: The one-liner creates quite a few allocations (zero allocations in the alternative) and it iterates over the array at least twice (one loop in the suggestion).

cpfiffer: > I am in disbelief that this code is more efficient than the above one-liner, but I checked with `@btime` and indeed... 🤯

If I had a nickle for each time I thought I had written the best code and then David has a better solution -- I'd be so rich

Issue #430, opened 2023-08-09
Overloading PosteriorStats methods


sethaxen:
As discussed on the Turing Slack, MCMCChains should overload methods defined in ArviZ's [PosteriorStats](https://github.com/arviz-devs/PosteriorStats.jl/) for `Chains` inputs, similar to how it overloads methods in MCMCDiagnosticTools.

This would introduce some breaking changes:
- `MCMCChains.summarize` would be replaced with `PosteriorStats.summarize`. The returned object would similarly be a table, but the API is different enough that this would be breaking.
- `MCMCChains.hpd` would be replaced with `PosteriorStats.hdi`.
  - ArviZ renamed `hpd` to `hdi` some time ago because it a strictly more descriptive name. The HDI is a distributional property, appropriate for both priors and posteriors, but also other distributions. A deprecation warning can be used here.
  - PosteriorStats also uses a default interval probability of 0.94. As McElreath argues, any default is arbitrary and using a "non-standard" probability encourages the user to think about the interval they select and avoids potential confusion for those coming from a more frequentist background. McElreath recommends 0.89 for no reason than because it's prime. Its estimates are also more stable with low sample size. We use 0.94 because it's more conservative.

This would introduce several new functions:
- `r2_score`
- `loo_pit`
- `loo`
- `waic`
- `compare`

A `loo` method is also defined and exported by ParetoSmooth, so if `loo` is also exported by MCMCChains, this would cause namespace collisions. We should discuss how we want to handle this.

cc @ParadaCarleton @devmotion

devmotion:
> A `loo` method is also defined and exported by ParetoSmooth, so if `loo` is also exported by MCMCChains, this would cause namespace collisions. We should discuss how we want to handle this.

```julia
julia> using MCMCChains

julia> :loo in names(MCMCChains; all=true, imported = true)
false
```

So `loo` is neither exported nor imported in MCMCChains.

sethaxen:
No, `loo` is exported by PosteriorStats. But actually, on second thought there'd be no benefit in overloading `loo` for `Chains`, since its input is an array of log-likelihood values that wouldn't be stored in `Chains` anyways.

PR #431, opened 2023-08-20
Import and extend PosteriorStats

sethaxen:
This PR makes the following replacements everywhere:
- `hpd` -> `PosteriorStats.hdi`
- `summarize` -> `PosteriorStats.summarize`
- `ChainDataFrame` -> `PosteriorStats.SummaryStats`

The replacement of `ChainDataFrame` and the slight change in API and behavior of the methods makes this a breaking change.

Implements #430

e.g.

```julia
julia> val = rand(500, 2, 3);

julia> chn
Chains MCMC chain (500×2×3 Array{Float64, 3}):

Iterations        = 1:1:500
Number of chains  = 3
Samples per chain = 500
parameters        = param_1, param_2


Summary Statistics
          mean    std    hdi_3%  hdi_97%  mcse_mean  mcse_std  ess_tail  ess_bulk  rhat 
 param_1  0.50  0.283  0.00745     0.929     0.0073    0.0034      1480      1531  1.00
 param_2  0.49  0.285  0.000140    0.926     0.0074    0.0034      1499      1506  1.00

Quantiles
            2.5%  25.0%  50.0%  75.0%  97.5% 
 param_1  0.0252  0.254  0.510  0.745  0.969
 param_2  0.0180  0.242  0.483  0.727  0.969

julia> hdi(chn)
HDI
             lower  upper 
 param_1  0.00745   0.929
 param_2  0.000140  0.926

julia> describe(chn)
2-element Vector{SummaryStats}:
 Summary Statistics (2 rows, 10 cols)
 Quantiles (2 rows, 6 cols)
```


src/stats.jl
```diff
@@ -119,28 +109,28 @@ function changerate(
...
     end
 end
 
-function chaindataframe_changerate(name, names_of_params, chains; kwargs...)
+function summarystats_changerate(name, names_of_params, chains; kwargs...)
     # Compute the change rates.
     changerates, mvchangerate = changerate(chains)
 
     # Summarize the results in a named tuple.
-    nt = (; zip(names_of_params, changerates)..., multivariate = mvchangerate)
```
sethaxen, line 140: Lacking a `parameter` column meant the `show` method was broken. But since there is a changerate for every parameter, it makes more sense to do the same thing as `gelmandiag_multivariate` and return a `SummaryStats` for the marginal values and return the multivariate changerate separately.

src/plot.jl
```diff
@@ -64,7 +64,7 @@ const supportedplots = push!(collect(keys(translationdict)), :mixeddensity, :cor
         lags = 0:(maxlag === nothing ? round(Int, 10 * log10(length(range(c)))) : maxlag)
         # Chains are already appended in `c` if desired, hence we use `append_chains=false`
         ac = autocor(c; sections = nothing, lags = lags, append_chains=false)
-        ac_mat = convert(Array, ac)
+        ac_mat = cat(reduce.(hcat, Iterators.drop.(ac, 1))...; dims=3)
```
torfjelde, line 67: How large is this `...` generally going to be? Could it cause a blow-up in compilation times?

torfjelde (approved):
I'm _very_ excited about this :eyes: 

I have a few minor comments, but will approve as it looks good:)

src/plot.jl
sethaxen: I think for many lags, `autocor` itself would actually have a problem, since `SummaryStats`, like `ChainsDataFrame`, is backed by a `NamedTuple` whose entries are the columns.

A safer approach would be to modify PosteriorStats so that `SummaryStats` wraps any Table. Then `autocor` could be modified to produce a `SummaryStats` backed by an `OrderedDict`. 

This `cat` call could be easily replaced by `Compat.stack`.

torfjelde: > I think for many lags, autocor itself would actually have a problem, since SummaryStats, like ChainsDataFrame, is backed by a NamedTuple whose entries are the columns.

Gotcha :+1: 

> This cat call could be easily replaced by Compat.stack.

Should we maybe do this regardless of whether `autcorr` suffers from the same issue? This way things will "just work" once we fix PosterioStats:)

sethaxen: I started on making `SummaryStats` wrap any Table and decided to make some breaking changes while I'm at it. Once I finish those, I'll update this PR accordingly.

torfjelde: Sounds good:)

devmotion:
It seemed you wanted to push a few additional updates @sethaxen? Or is this PR ready from your side?

sethaxen:
Yeah it's pending a breaking changes to PosteriorStats I need to finish up, to avoid another breaking release here.

sethaxen:
This is ready for another review.

sethaxen:
Currently PosteriorStats has no `try/catch` mechanism for if a given statistic fails. That causes chains with less than 10 draws to error upon computation of summary stats (or display). For PosteriorStats that makes sense, but it's a major nuisance for MCMCChains. I actually think this has come up enough that it makes sense for MCMCDiagnosticTools to raise an informative warning and return `NaN`s in such cases. Will open an issue there.

devmotion:
Oh, I missed https://github.com/TuringLang/MCMCChains.jl/pull/431#issuecomment-1868452766, probably due to the Christmas break. I added the PR to my todo list 🙂

src/MCMCChains.jl
```diff
@@ -47,7 +47,11 @@ export mcse
 export rafterydiag
 export rstar
 
-export hpd
+# Reexport stats functions
+using PosteriorStats: SummaryStats, default_diagnostics, default_stats,
+    default_summary_stats, hdi, summarize
+export SummaryStats, default_diagnostics, default_stats, default_summary_stats, hdi,
+    summarize
```
devmotion, line 54: Should we actually reexport all of these functions? Based on the tests it seems
```suggestion
export SummaryStats, hdi, summarize
```
might be sufficient?

src/chains.jl
```diff
@@ -346,11 +346,14 @@ function Base.show(io::IO, chains::Chains)
 end
 
 function Base.show(io::IO, mime::MIME"text/plain", chains::Chains)
-    print(io, "Chains ", chains, ":\n\n", header(chains))
+    println(io, "Chains ", chains, ":\n\n", header(chains))
```
devmotion, line 349: Is codecov right? Do we not test this function?

src/plot.jl
```diff
@@ -64,7 +64,7 @@ const supportedplots = push!(collect(keys(translationdict)), :mixeddensity, :cor
         lags = 0:(maxlag === nothing ? round(Int, 10 * log10(length(range(c)))) : maxlag)
         # Chains are already appended in `c` if desired, hence we use `append_chains=false`
         ac = autocor(c; sections = nothing, lags = lags, append_chains=false)
-        ac_mat = convert(Array, ac)
+        ac_mat = stack(map(stack ∘ Base.Fix2(Iterators.drop, 1), ac))
```
devmotion, line 67: This seems a bit inefficient (particular the use of `stack(map(stack`) - maybe we should preallocate the desired array and copy directly? Possibly a utility function could be added to PosteriorStats?

src/plot.jl
```diff
@@ -195,37 +195,37 @@ function _compute_plot_data(
...
-    show_hpdi = true,
+    show_hdii = true,
     fill_q = true,
-    fill_hpd = false,
+    fill_hdi = false,
     ordered = false
 )
 
-    chain_dic = Dict(zip(quantile(chains)[:,1], quantile(chains)[:,4]))
+    chain_dic = Dict(zip(quantile(chains)[2], quantile(chains)[5]))
```
devmotion, line 212: Maybe a good opportunity to make the code a bit more efficient:
```suggestion
    quantile_chains = quantile(chains)
    chain_dic = Dict(zip(quantile_chains[2], quantile_chains[5]))
```
(it would also be nice if it would be possible to use something more descriptive than `2` and `5`)

src/MCMCChains.jl
sethaxen: It depends on how accessible we want these methods to be, as they offer different summary presets that I think will be useful to users. I also expect most users won't even know PosteriorStats exists so wouldn't know they exist otherwise. But I could also see the argument in leaving them out.

src/plot.jl
sethaxen: It may not be the most efficient, but compared to the actual `autocor` call, it should not be the computational bottleneck, and not using `stack` requires substantially more code.

It might make sense to add to PosteriorStats a function that converts a `SummaryStats` to a matrix, but this would effectively do the same thing as the inner `stack`. What kind of utility function do you have in mind?

devmotion: > What kind of utility function do you have in mind?

Basically a `convert`, `Array`/`Matrix` constructor`, or other function that performs the same thing (if the first two alternatives are not desirable). So that downstream packages (such as MCMCChains) don't have to deal with internals of `SummaryStats`.

src/plot.jl
sethaxen: Hmm, yeah this and the code immediately below are a bit of a mess. Will rework to use the new methods.

src/plot.jl
sethaxen: This line is not using any internals of `SummaryStats`, whose docstring explains it is an OrderedDict-like Table that can be iterated/indexed by its column names. 

I think the methods we're missing in PosteriorStats are
```julia
function Base.getindex(stats::SummaryStats, cols::Union{Colon,AbstractVector{Int},AbstractVector{Symbol}})
    cols isa Colon && return Tables.matrix(stats)
    return stack(Tables.getcolumn(stats, k) for k in cols)
end
Base.firstindex(s::SummaryStats) = 1
Base.lastindex(s::SummaryStats) = length(s)
```

Then this would be
```julia
stack(ac_i[2:end] for ac_i in ac)
```
but this is no more efficient.

src/stats.jl
```diff
@@ -185,57 +200,43 @@ function describe(
...
+    stats = [summarystats(chains; kwargs...), quantile(chains; q = q, kwargs...)]
+    return stats
 end
 
 """
-    hpd(chn::Chains; alpha::Real=0.05, kwargs...)
+    hdi(chn::Chains; prob::Real=0.94, kwargs...)
 
-Return the highest posterior density interval representing `1-alpha` probability mass.
+Return the unimodal highest density interval (HDI) representing `prob` probability mass.
```
torfjelde, line 212: What's the meaning of "unimodal" here?

src/stats.jl
torfjelde: Genuine question: what is the change-rate in this context?

src/stats.jl
```diff
@@ -17,24 +17,43 @@ Setting `append_chains=false` will return a vector of dataframes containing the
...
-        chains, funs...;
-        func_names = func_names,
-        append_chains = append_chains,
-        name = "Autocorrelation",
-        kwargs...
-    )
+    # Obtain names of parameters.
+    names_of_params = var_names === nothing ? names(chn) : var_names
+
+    # set up the functions to be evaluated
```
torfjelde, line 32: Maybe not use the phrase `functions` here? `lags` are just integers right? 
Maybe
```suggestion
    # Construct column names for lags.
```

src/stats.jl
```diff
@@ -17,24 +17,43 @@ Setting `append_chains=false` will return a vector of dataframes containing the
...
+    col_names = Symbol.("lag", lags)
+
+    # avoids using summarize directly to support simultaneously computing a large number of
+    # lags without constructing a huge NamedTuple
+    if append_chains
+        # Evaluate the functions.
```
torfjelde, line 38: And here
```suggestion
```

src/stats.jl
```diff
@@ -17,24 +17,43 @@ Setting `append_chains=false` will return a vector of dataframes containing the
...
+        data = _permutedims_diagnostics(chn.value.data)
+        vals = stack(map(eachslice(data; dims=3)) do x
+            return autocor(vec(x), lags; demean=demean)
+        end)
+        table = Tables.table(vals'; header=col_names)
+        return SummaryStats("Autocorrelation", table, names_of_params)
+    else
+        # Evaluate the functions.
```
torfjelde, line 46:
```suggestion
```

torfjelde (commented):
Added a few comments!

Am very excited about this though; bloody great stuff @sethaxen :heart:

src/stats.jl
sethaxen: The default estimator used by `hdi` assumes that the distribution is unimodal. The alternative (not yet in PosteriorStats but doable with HighestDensityRegions.jl) for multimodal distributions first fits a KDE and then partitions into one or more intervals by density. For fast summary statistics, the unimodal version is more useful.

src/stats.jl
sethaxen: From inspecting the code, it's, for each parameter and chain, the fraction of draws that are different from the previous draw. I suppose it's similar to "acceptance rate."

src/plot.jl
devmotion: > This line is not using any internals of SummaryStats, whose docstring explains it is an OrderedDict-like Table that can be iterated/indexed by its column names.

I was referring to the `2:end` which seems a bit special. I missed that the docstring mentions that the first column is reserved for parameter names.

src/plot.jl
sethaxen: Yeah it was able to be simplified a lot. But also, for n parameters, it was computing the median n^2 times. This has now been fixed.

src/discretediag.jl
```diff
@@ -17,16 +17,16 @@ function MCMCDiagnosticTools.discretediag(
...
+        "Chisq diagnostic - Between chains", between_chain_vals, param_names,
     )
-    within_chain_dfs = map(1:size(_chains, 3)) do i
+    within_chain_stats = map(1:size(_chains, 3)) do i
         vals = map(val -> val[:, i], within_chain_vals)
-        return ChainDataFrame("Chisq diagnostic - Chain $i", merge(parameters, vals))
+        return SummaryStats("Chisq diagnostic - Chain $i", vals, param_names)
     end
-    dfs = vcat(between_chain_df, within_chain_dfs)
+    stats = [between_chain_stats, within_chain_stats...]
```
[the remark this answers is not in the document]
sethaxen: Thanks, I went with the 2nd option.

src/chains.jl
sethaxen: Yes, it seems there was no test. I added one.

sethaxen:
This is ready for another review. Also, it would be nice to get input on https://github.com/arviz-devs/PosteriorStats.jl/issues/25, since that's relevant for this PR.

devmotion:
I thought https://github.com/arviz-devs/PosteriorStats.jl/issues/25 holds back this PR?

sethaxen:
@devmotion is right, I'd like to merge that breaking change and also some breaking changes to make the `SummaryStats` object easier to interact with (and a little more like `ChainDataFrame`). Otherwise when I make those changes to PosteriorStats, it would require another major version bump here and a minor one for Turing.

After a long break I'm back to working on arviz PRs so I should be able to wrap that up in the next couple of weeks.

yebai:
@sethaxen, just wondering what is missing from this PR?

sethaxen:
> @sethaxen, just wondering what is missing from this PR?
Sorry @yebai for the delay in finishing this. arviz in both Python and Julia have been undergoing some API changes as Python arviz pushes to 1.0, so I've been holding off on a breaking release to PosteriorStats followed by an update to this PR until things stabilize. But since it's been taking a while I think I'll just make the release in the next few weeks and update here, with the understanding that the PosteriorStats API may shortly change again (which could require a breaking release here as well).

src/plot.jl
```diff
@@ -47,7 +47,11 @@ The following options are available:
 
 - `q` (default: `[0.1, 0.9]`): The two quantiles used for plotting if `fill_q = true` or `show_qi = true`.
 
-- `hpd_val` (default: `[0.05, 0.2]`): The complementary probability mass(es) of the highest posterior density intervals that are plotted if `fill_hpd = true` or `show_hpdi = true`.
+- `ci_fun` (default: `eti`): The function used to compute the credible intervals.
+  (Can be [`eti`](@ref) or [`hdi`](@ref))
+
+- `ci_probs` (default: `[$DEFAULT_CI_PROB, 0.8]`): The probability mass(es) of the credible
```
sethaxen, line 114: These are now quite close; perhaps the 0.8 should be decreased. For reference, ArviZ uses `[0.89, 0.5]` as the default probs.

penelopeysm: That makes sense, happy to change it to 0.5.

src/summarize.jl
```diff
@@ -1,137 +1,24 @@
...
 
-# Index functions
-function Base.getindex(c::ChainDataFrame, s::Union{Colon, Integer, UnitRange}, g::Union{Colon, Integer, UnitRange})
-    convert(Array, getindex(c, c.nt[:parameters][s], collect(keys(c.nt))[g]))
-end
+`stats_funs` is a collection of functions that reduces a matrix with shape `(draws, chains)`
+to a scalar or a collection of scalars. Alternatively, an item in `stats_funs` may be a
+`Pair` of the form `name => fun` specifying the name to be used for the statistic or of the
+form `(name1, ...) => fun` when the function returns a collection. When the function returns
+a collection, the names in this latter format must be provided.
```
penelopeysm, line 15: Would it be possible to add a sentence to the docstring saying what `append_chains` and `name` are?

sethaxen: Done! I opted to add bullet points listing keyword arguments and only document those that MCMCChains modifies somehow (for the rest, the user is referred to the PosteriorStats docs)

penelopeysm:
Thanks for the explanation!

Just the disclaimer to begin, I have absolutely no intention of reviewing this unfairly, but I'm also cognisant that there's some level of conflict of interest because I'm also separately working on FlexiChains (and trying to solve all these same tricky decisions with APIs!), and I can't prove that I'm fully disinterested. Feel free to say if you'd rather someone else review, I won't be offended in the least.

Whilst I'm very happy with the general aim of centralising functionality and reusing code, I think as it stands it's a loss in terms of usability if it's not possible to use the parameter name somehow. IMO, I should be able to combine the three things `ss`, `:mean`, and `:a` in order to get the mean of `:a` stored in `ss`. `ChainDataFrame` does this:

```julia
julia> using MCMCChains; ss = mean(Chains(rand(3, 2), [:a, :b]))
Mean
  parameters      mean
      Symbol   Float64

           a    0.6386
           b    0.5754


julia> ss[:a, :mean]
0.6385968125591274
```

For `SummaryStats` right now I've only come up with this, which is a bit verbose:

```julia
julia> using MCMCChains; ss = mean(Chains(rand(3, 2), [:a, :b]))
Mean
     mean
 a  0.460
 b  0.647

julia> ss[:mean][findfirst(isequal(:a), ss[:label])]
0.46036692554959524
```

Once there is an interface like that, I'd be very happy to approve this PR. I'm not particularly wedded to the interface being `getindex` or `ss[:a, :mean]`, or row-based indexing in general. For example if you had `ss[:mean]` return a NamedTuple of `(a = 0.460, b = 0.647)` then we could do `ss[:mean].a` or `ss[:mean][:a]`, and I'd be fine with that too (as long as it's documented -- which brings us to the next paragraph).

Regarding documentation in general: I do indeed see that you're documenting things in PosteriorStats.jl (which is great!) but if we bring PosteriorStats into MCMCChains, we also need to worry about it in MCMCChains and (perhaps more importantly) the main Turing docs because this is very user-facing. Currently the interface of MCMCChains is quite underdocumented (it essentially amounts to 'learn by reading the examples in the docs'). This existing situation is of course not your fault at all, but I'd like to make sure that when we add new functionality, someone also ensures that it's explained in the main Turing docs.

sethaxen:
To clarify, no changes would be needed at all to MCMCChains to support a new indexing syntax for `SummaryStats`; all changes would be in PosteriorStats, and they would be non-breaking. While prioritizing tasks, I had placed swapping PosteriorStats into MCMCChains as higher priority than finalizing `SummaryStats`'s interface, but I can see @penelopeysm's reasonable point that the priorities could be reversed, and I'm fine with holding off on merging this PR until that interface is finalized. I'll take a pass at a design issue and will link it here for feedback.

PR #440, opened 2023-10-26
Update Project.toml

yebai:


devmotion:
AFAICT this requirement in the registry was temporarily reverted due to issues with BinaryBuilder: https://github.com/JuliaRegistries/General/pull/94185 So I think it would be better to wait with updating compat entries until everything is fully in place.

devmotion:
The new MCMCChains version is released: https://github.com/JuliaRegistries/General/pull/94177

PR #454, opened 2024-03-28
Add plotting via Makie as Extension

alecloudenback:
## Current Status:

A color-blind-friendly color scheme (`:tol_vibrant`) is used to plot the trance and parameter densities for each parameter in each chain.

Sample plots:

**Basic Layout**
image

**Large number of parameters**
image

** Large number of chains**
image

## To-do

- [ ] Documentation
- [ ] Tests (test that plot gets created for a variety of sizes?)
- [ ] ~~Extend `Makie.plot` instead of a new `myplot` function~~ ([not good idea](https://github.com/TuringLang/MCMCChains.jl/pull/454#issuecomment-2028421641))
- [ ] Gracefully handle more than 7 chains for the color scheme

## Sample Script

```julia
julia> using MCMCChains, CairoMakie
julia> chns = Chains(randn(300, 5, 3), [:A, :B, :C, :D, :E]);
julia> myplot(chns)
```


devmotion (commented):
I think this approach is a bit too opinionated and inflexible. Could we instead define `convert_arguments` (to tell Makie how it should view and use chains) and - if a new type of plots is needed - add a plot recipe (see https://docs.makie.org/dev/explanations/recipes/)?

alecloudenback:
After talking with the Makie devs on Discord:

- The `convert_arguments` isn't really a path because all it will do is be able to plot the data into a single plot (no faceting by parameter or chain) or convey metadata (e.g. parameter name or chain number)
- Recipes are for creating a single sub-plot, not defining a figure with multiple sub-plots as desired
- The recommendation was to just define a new function that would plot as desired.

The latest commit is the recommended approach, including [registering an error hint](https://github.com/jkrumbiegel/MakiePkgExtTest/blob/c1718640f96ba0dc72d2baaa489a1fb1097bfbc5/SomePackage/src/SomePackage.jl#L27) if a Makie-related package isn't yet loaded and `trace` is called.

alecloudenback:
@devmotion / @theogf pinging since you had some initial feedback requesting a recipes-based approach, but I don't think that works when wanted to plot multiple chains/parameters (see my last post). If good with the overall approach here, I can knock out the remaining To-Dos.

ext/MCMCChainsMakieExt.jl
```diff
@@ -0,0 +1,74 @@
+module MCMCChainsMakieExt
+
+import MCMCChains
+import Makie
+
+
+function MCMCChains.trace(chns::T) where {T<:MCMCChains.Chains}
```
theogf, line 7: I don't think `trace` is an appropriate name, it should be something like `traceplot` or `plot_chains` to make it clear this is outputting a figure.

If you decide to go the method dispatch way, it would be good to deconstruct it a bit better:
- Make a mutable function that takes an `Axis`
- Create separate functions for each subplot.

This way the code is beneficial for anyone who would just like some part of the code.

ext/MCMCChainsMakieExt.jl
```diff
@@ -0,0 +1,74 @@
...
+    params = MCMCChains.names(chns, :parameters)
+
+    n_chains = length(MCMCChains.chains(chns))
+    n_samples = length(chns)
+    n_params = length(params)
+
+    colors = Makie.to_colormap(:tol_vibrant)
```
theogf, line 14: As @devmotion this is a bit too opiniated, what if I want to use a different set of colors?

ext/MCMCChainsMakieExt.jl
```diff
@@ -0,0 +1,74 @@
...
+    width = 600
+    height = max(400, 80 * n_params)
```
theogf, line 16: Same for this, it is too restrictive...

theogf (changes requested):
The plots you are proposing look good, but the problem is that it is too restrictive/opiniated for being in a package like MCMCChains.jl. If `convert_arguments` is not sufficient, you can use the `@recipe` macro no?
https://docs.makie.org/stable/explanations/recipes/#full_recipes_with_the_recipe_macro

ext/MCMCChainsMakieExt.jl
alecloudenback: > I don't think trace is an appropriate name, it should be something like traceplot or plot_chains to make it clear this is outputting a figure.

`trace` follows the convention that plot names in the Julia ecosystem don't have the `plot`, e.g. `scatter`, `lines, etc.

>If convert_arguments is not sufficient, you can use the @recipe macro no?

Recipes are for creating a single sub-plot, not defining a figure with multiple sub-plots. I'm not sure there's a way to use a recipe unless all chains and parameters were plotted into a single sub-plot and not faceted the way MCMC plots tend to be

> Make a mutable function that takes an Axis

👍 

> Create separate functions for each subplot.

i.e. where you could get the left side and right sight separately?

ext/MCMCChainsMakieExt.jl
alecloudenback: I will explore ways to make more flexible.

yebai:
See also, https://github.com/simonsteiger/ChainsMakie.jl

yebai:
@alecloudenback, what is missing for this PR to be completed?

alecloudenback:
See https://github.com/TuringLang/MCMCChains.jl/pull/454#issuecomment-2028421641

The way that Makie wants to have multiple sub-plots isn't conducive to being flexible with user-facing plot config (the `@recipe`/`convert_arguments` approach). The inflexible way is the way that I created but the feedback was that it was too opinionated and would not allow users to customize it easily. I don't really know how to resolve it.

devmotion:
Given that chains support the Tables interface, maybe users should be pointed towards AlgebraOfGraphics as the easiest (but still customizable) way for plotting chains with Makie?

yebai:
Closed in favour of https://github.com/simonsteiger/ChainsMakie.jl

PR #460, opened 2024-08-29
Revert "Add documentation and docstrings for ridgeline, forest and caterpilar plots"

devmotion:
Reverts TuringLang/MCMCChains.jl#328

The PR broke the documentation build (it refers to non-existing docstrings and included docstrings in the markdown files). Ideally, these issues should be fixed such that the additional docs are not lost but I'm currently on my phone so I cannot make major changes easily.

PR #461, opened 2024-08-29
Move docstrings to respective functions

devmotion:
Alternative to #460 that hopefully fixes the failing docs build on the master branch.


devmotion:
OK, much better. Now the only missing piece is that Documenter complains because `forestplot` and `ridgelineplot` (and the inplace versions) have a docstring but are not included in the docs.

devmotion:
Documentation seems to build successfully again: https://turinglang.org/MCMCChains.jl/previews/PR461/statsplots/

The test failure on Windows seems weird but unrelated.

Issue #443, opened 2023-11-07
names() and get() for a specific section should always return parameters in the same order


bgctw:
I encountered the following behaviour in a complex Chains object inspected from Infiltrator prompt,
where names and get return variables in a different order (similar to #326):
```
infil> names(df.chn, :u0)                                                                                                                                                                    
11-element Vector{Symbol}:                                                                                                                                                                   
 :s₊α_R0                                                                                                                                                                                     
 :s₊R_N0                                                                                                                                                                                     
 :s₊I_P0                                                                                                                                                                                     
 :s₊R_P0                                                                                                                                                                                     
 :s₊L_P0                                                                                                                                                                                     
 :s₊α_P0                                                                                                                                                                                     
 :s₊L0                                                                                                                                                                                       
 :s₊L_N0                                                                                                                                                                                     
 :s₊I_N0                                                                                                                                                                                     
 :s₊R0                                                                                                                                                                                       
 :s₊B0                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                                                                                       
infil> keys(get(df.chn, section=:u0))                                                                                                                                                                                                                                                                                                                                                   
(:s₊R_N0, :s₊R_P0, :s₊L_N0, :s₊I_N0, :s₊L0, :s₊α_R0, :s₊B0, :s₊L_P0, :s₊α_P0, :s₊R0, :s₊I_P0)   
```

I failed to construct a MWE. The following works as expected:

```
ulia> chn1 = Chains(rand(100, 4, 2), [:a, :b, :c, :d]);

julia> chn1s = set_section(chn1, Dict(:parameters => [:a,:b], :u0 => [:d, :c]));

julia> names(chn1s, :u0)
2-element Vector{Symbol}:
 :d
 :c

julia> keys(get(chn1s, section = :u0))
(:d, :c)
```

Encountered on Julia 1.9.2 and MCMCChains v6.0.4

penelopeysm:
Closed by #466

PR #466, opened 2024-12-04
keep parameter order

a1ix2:
Addresses https://github.com/TuringLang/MCMCChains.jl/issues/443

### DIFF
--- a/src/chains.jl
+++ b/src/chains.jl
@@ -240,7 +240,7 @@ julia> get(chn, :param_1; flatten=true)
 ```
 """
 function Base.get(c::Chains, vs::Vector{Symbol}; flatten=false)
-    pairs = Dict()
+    pairs = OrderedCollections.OrderedDict()
     for v in vs
         syms = namesingroup(c, v)
         len = length(syms)
@@ -289,7 +289,7 @@ function Base.get(
     section::Union{Symbol,AbstractVector{Symbol}},
     flatten = false
 )
-    names = Set(Symbol[])
+    names = OrderedCollections.OrderedSet(Symbol[])
     regex = r"[^\[]*"
     _section = section isa Symbol ? (section,) : section
     for v in _section
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -56,7 +56,7 @@ function cummean(x::AbstractVector)
     return y
 end
 
-function _dict2namedtuple(d::Dict)
+function _dict2namedtuple(d::AbstractDict)
     t_keys = ntuple(x -> Symbol(collect(keys(d))[x]), length(keys(d)))
     t_vals = ntuple(x -> collect(values(d))[x], length(values(d)))
     return NamedTuple{t_keys}(t_vals)
--- a/test/sections_tests.jl
+++ b/test/sections_tests.jl
@@ -1,9 +1,12 @@
 using  MCMCChains, Test
 
 # https://github.com/TuringLang/AdvancedMH.jl/pull/63
+# https://github.com/TuringLang/MCMCChains.jl/issues/443
 @testset "order of parameters" begin
-    chains = Chains(rand(2, 10, 4), vcat(["μ[$i]" for i in 1:9], :lp), (internals=[:lp],))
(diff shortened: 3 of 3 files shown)


mhauru:
Looks good to me, thanks @a1ix2! Did you run into a case where this caused you trouble? If so, do you have a reproducible example that triggers the issue (even non-minimal)? Would be good to have a test for this too, but the original issue you reference didn't come with an MWE.

a1ix2:
For me it was mostly an annoyance when trying to interoperate with `ComponentArrays` when using `get`. Here's a more or less useless example but still.

```julia
using Turing, Distributions

@model function fakemodel(xs)
    mu1 ~ Normal(10.0)
    mu2 ~ Normal(10.0)
    sigma1 ~ InverseGamma(2.0, 1.0)
    sigma2 ~ InverseGamma(2.0, 1.0)
    for i in eachindex(xs)
        xs[i] ~ MixtureModel([Normal(mu1, sigma1), Normal(mu2, sigma2)], [0.5, 0.5])
    end
end

xs = rand(MixtureModel([Normal(0.1, 1.0), Normal(3.0, 0.5)], [0.5, 0.5]), 1000)

model = fakemodel(xs)

chain = sample(model, NUTS(), 100)

all_chains = get_params(chain)
keys(allchains)
# which returns an NTuple with a jumble of unordered internal and parameters
# (:hamiltonian_energy, :n_steps, :numerical_error, :max_hamiltonian_energy_error, :mu2, :sigma1, :hamiltonian_energy_error, :mu1, :is_accept, :sigma2, :log_density, :tree_depth, :step_size, :acceptance_rate, :lp, :nom_step_size)

all_param_chains = get(chain, section=:parameters)
# Also returns a NamedTuple with unordered parameters
# (mu1 = [0.16170966805695142; 0.1554824980047796; … ; 0.08014927911102815; 0.0922168468995448;;], mu2 = [2.9505553517683385; 3.0352436810243066; … ; 2.9597771408952247; 2.9428157073776444;;], sigma2 = [0.549196145152957; 0.5526604135099159; … ; 0.5646327205926919; 0.6056774396897618;;], sigma1 = [0.8824624114147762; 0.9787977609583128; … ; 0.9162592661201038; 0.931760395213333;;])

keys(all_param_chains)
# (:mu1, :mu2, :sigma2, :sigma1)
```

Ideally I'd want to obtain a chain as a vector of ComponentArrays, one for each iteration, so that I can splat more easily into other blackbox optimizers like CMAEvolutionStrategy and so on. Right now I have to rely on writing a return statement in the model with all parameters manually listed

```julia
using ComponentArrays: ComponentArray

@model function fakemodel_ret(xs)
    mu1 ~ Normal(10.0)
    mu2 ~ Normal(10.0)
    sigma1 ~ InverseGamma(2.0, 1.0)
    sigma2 ~ InverseGamma(2.0, 1.0)
    for i in eachindex(xs)
        xs[i] ~ MixtureModel([Normal(mu1, sigma1), Normal(mu2, sigma2)], [0.5, 0.5])
    end
    
    return ComponentArray(mu1=mu1, mu2=mu2, sigma1=sigma1, sigma2=sigma2
end

model_ret = fakemodel_ret(xs)

chain_ret = sample(model_ret, NUTS(), 100)

qs = generated_quantities(model_ret, chain_ret)
```

or if I want a flat vector of the best chain samples I use something like
```
_bestidx = findmax(chain.value[var=:log_density])[2][1]
_bestsample = chain.value[_bestidx, 1:end-12, 1]
```
where the `end-12` skips the 12 internals (I _think_ this number is dependent on the particular sampler?), but this gives me an AxisArray I can't easily cast into a ComponentArray (which I find much more intuitive to deal with).

While I haven't yet figured out a way to cleanly do these kinds of things using MCMCChains.jl this was the first

mhauru (approved):
I added a test that covers the changes here. Thanks @a1ix2, much appreciated!

radka-j (approved):
LGTM :)

mhauru:
Live once [v6.0.7](https://github.com/JuliaRegistries/General/pull/120833) is out.

PR #471, opened 2025-02-09
Documentation and Turing Navigation CI improvement

shravanngoswamii:
Using [Turing Actions](https://github.com/TuringLang/actions) for Documentation and global Turing Navigation bar!
Reference: https://github.com/TuringLang/actions/issues/12

### DIFF
--- a/.github/workflows/Docs.yml
+++ b/.github/workflows/Docs.yml
@@ -24,3 +24,5 @@ jobs:
     steps:
       - name: Build and deploy Documenter.jl docs
         uses: TuringLang/actions/DocsDocumenter@main
+        env:
+          GKSwstype: nul
--- a/.github/workflows/Docs.yml
+++ b/.github/workflows/Docs.yml
@@ -1,4 +1,4 @@
-name: Docs
+name: Documentation
 
 on:
   push:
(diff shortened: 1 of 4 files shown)


.github/workflows/Docs.yml
```diff
@@ -1,29 +1,28 @@
...
-        with:
-          version: '1'
-      - name: Install dependencies
-        run: julia --project=docs/ -e 'using Pkg; Pkg.develop(PackageSpec(path=pwd())); Pkg.instantiate()'
-      - name: Build and deploy
-        run: julia --project=docs docs/make.jl
+      - name: Build and deploy Documenter.jl docs
+        uses: TuringLang/actions/DocsDocumenter@main
         env:
           GKSwstype: nul
```
penelopeysm, line 28:
```suggestion
```

Might not be needed anymore since we removed it on AdvancedPS and the docs built fine? Let's see what happens here

penelopeysm (changes requested):

penelopeysm:
CI is failing because of https://github.com/MakieOrg/Makie.jl/issues/4759

penelopeysm (approved):

.github/workflows/Docs.yml
devmotion: This environment variable tells Plots.jl (or rather the standard GR backend) to operate in headless mode which avoids errors and warnings in CI. Maybe there were some recent improvements such that it is not as important anymore nowadays. It's still used in CI of GR.jl itself though: https://github.com/jheinen/GR.jl/blob/857e8d158d8c7cdeacbd34dd7ddac1e3deeb1dcf/.github/workflows/ci.yml#L13 (together with https://github.com/jheinen/GR.jl/blob/857e8d158d8c7cdeacbd34dd7ddac1e3deeb1dcf/.github/workflows/ci.yml#L12 which has been important for correctly printing unicode labels and is mentioned also in e.g. a few issues and discussions in MCMCChains)

PR #476, opened 2025-02-26
1.6.0 CI fails, need to bump minimum version

penelopeysm:
See e.g. https://github.com/TuringLang/MCMCChains.jl/actions/runs/13531292096/job/37813844525

```
ERROR: LoadError: LoadError: InitError: could not load library "/home/runner/.julia/artifacts/8a8d0ba819ce1e30b93f34cc61f891d1334be59d/lib/libQt6Concurrent.so"
/opt/hostedtoolcache/julia/1.6.0/x64/bin/../lib/julia/libstdc++.so.6: version `GLIBCXX_3.4.28' not found (required by /home/runner/.julia/artifacts/8a8d0ba819ce1e30b93f34cc61f891d1334be59d/lib/libQt6Core.so.6)
```

This is because of https://github.com/JuliaLang/julia/issues/40585 which was fixed in https://github.com/JuliaLang/julia/pull/41645 and backported to 1.6.3

(This wasn't caught previously because we were running CI with the julia version set to `'1.6'`, which resolves to the latest patch version of 1.6. In #474 I changed the version to `'min'`, which resolves to the earliest possible version specified by the julia compat entry in Project.toml, and then this issue appeared.)


mhauru (approved):
Good stuff, thanks Penny

PR #494, opened 2025-10-03
BFMI function

### LINKED ISSUE #484: Add function to calculate BFMI
Calculate the estimated Bayesian fraction of missing information (BFMI).

ref: https://python.arviz.org/en/latest/api/generated/arviz.bfmi.html#arviz.bfmi

For more information on BFMI, see https://arxiv.org/pdf/1604.00695v1.pdf

evandieren:
Adding the BFMI overlead from `MCMCDiagnosticTools.jl` to `MCMCChains.jl`

Exported `bfmi` and added tests.

Closes #484

### DIFF
--- a/test/diagnostic_tests.jl
+++ b/test/diagnostic_tests.jl
@@ -30,7 +30,8 @@ chn_disc = Chains(val_disc, start = 1, thin = 2)
 # Chains object using Hamiltonian energy for bfmi test
 chn_bfmi = Chains(rand(100, 4, 2), [:a, :b, :c, :hamiltonian_energy])
 chn_bfmi = set_section(
-    chn_bfmi, Dict(:internals => [:hamiltonian_energy], :parameters => [:a, :b, :c])
+    chn_bfmi,
+    Dict(:internals => [:hamiltonian_energy], :parameters => [:a, :b, :c]),
 )
 
 @testset "basic chains functions" begin
@@ -242,12 +243,12 @@ end
     # test bfmi function
     bfmi_values = bfmi(chn_bfmi)
     @test isa(bfmi_values, Vector{Float64})
-    @test size(bfmi_values, 1) == 2 
+    @test size(bfmi_values, 1) == 2
 end
 
 
 @testset "vector of vectors" begin
-    val = [rand(20) for _ in 1:10]
+    val = [rand(20) for _ = 1:10]
 
     chn = Chains(val)
     chn2 = Chains(reduce(hcat, val)')


src/stats.jl
```diff
@@ -391,3 +391,27 @@ end
...
+Compute the Bayesian Fraction of Missing Information (BFMI) for a set of chains.
+
+This is a diagnostic for Hamiltonian Monte Carlo (HMC) methods, which indicates how well the
+    momentum resampling explores the energy distribution.
+
+Returns a Vector{Float64} of BFMI values, one for each chain.
+
+Remark: This function is a wrapper around the `MCMCDiagnosticTools.jl` function.
+"""
+function bfmi(chains::Chains)
```
sethaxen: The trouble with exporting a `bfmi` function instead of overloading `MCMCDiagnosticTools.bfmi` is that if a user has both packages in scope then the user needs to explicitly prefix any call to `bfmi` with the package name since they collide.
```suggestion
function MCMCDiagnosticTools.bfmi(chains::Chains)
```

evandieren: Thank you for the review @sethaxen, I have made the changes in the commit I just added.

sunxd3:
@evandieren thanks for this. It looks good to me, once the [formatting](https://github.com/TuringLang/MCMCChains.jl/actions/runs/18257981270/job/51989067701?pr=494) is fixed, I think it's ready

src/stats.jl
```diff
@@ -391,3 +391,25 @@ end
...
+
+
+
+"""
+    MCMCDiagnosticTools.bfmi(chains::Chains)
+
+Compute the Bayesian Fraction of Missing Information (BFMI) for a set of chains.
+
+This is a diagnostic for Hamiltonian Monte Carlo (HMC) methods, which indicates how well the
+    momentum resampling explores the energy distribution.
```
mhauru:
```suggestion
This is a diagnostic for Hamiltonian Monte Carlo (HMC) methods, which indicates how well the momentum resampling explores the energy distribution.
```

src/stats.jl
```diff
@@ -391,3 +391,25 @@ end
...
+
+Returns a Vector{Float64} of BFMI values, one for each chain.
+"""
+function MCMCDiagnosticTools.bfmi(chains::Chains)
+
+    if :hamiltonian_energy ∉ names(chains)
+        throw(ArgumentError("The chain does not contain the `:hamiltonian_energy` parameter.
```
mhauru: These sorts of long strings can be a bit tricky to make fit into line length limits. There are a few ways to do it, my favourite is using tripe-quoted strings: https://docs.julialang.org/en/v1/manual/strings/#Triple-Quoted-String-Literals

mhauru (commented):
I agree with @sunxd3, all good except for formatting. That should be fixable by just running JuliaFormatter.jl (v1, not v2!) on your local branch. We've got some instructions [here](https://turinglang.org/docs/developers/contributing/#installing-juliaformatter), happy to clarify if need be.

mhauru (commented):
I'm happy to approve once CI passes.

PR #498, opened 2025-10-15
Add compatibility with PrettyTable v3.0

gragusa:
MCMCChains currently targets the PrettyTables.jl 2.x API. This PR adds compatibility with PrettyTables.jl 3.x while preserving the existing 2.x behavior. Downstream packages that still depend on PrettyTables 2.x will continue to work, and environments that can upgrade to 3.x will automatically use the new API.

### DIFF
--- a/src/summarize.jl
+++ b/src/summarize.jl
@@ -31,11 +31,9 @@ function Base.show(io::IO, ::MIME"text/plain", df::ChainDataFrame)
         println(io, df.name)
         # Support for PrettyTables 0.9 (`borderless`) and 0.10 (`tf_borderless`)
         PrettyTables.pretty_table(
-            io,
-            df.nt;
+            io, df.nt;
             formatters = formatter,
-            tf = isdefined(PrettyTables, :borderless) ? PrettyTables.borderless :
-                 PrettyTables.tf_borderless,
+            tf = isdefined(PrettyTables, :borderless) ? PrettyTables.borderless : PrettyTables.tf_borderless,
         )
     else
         # v3: ft_printf  ->  fmt__printf  (returns a formatter function)
@@ -46,12 +44,11 @@ function Base.show(io::IO, ::MIME"text/plain", df::ChainDataFrame)
         # v3: `tf` keyword was replaced by `table_format`
         #     use the predefined border set "borderless"
         PrettyTables.pretty_table(
-            io,
-            df.nt;
+            io, df.nt;
             backend = :text,  # explicit; :text is the default
             formatters = [fmt],  # v3 expects a Vector{Function}
             table_format = PrettyTables.TextTableFormat(
-                borders = PrettyTables.text_table_borders__borderless,
+                borders = PrettyTables.text_table_borders__borderless
             ),
         )
     end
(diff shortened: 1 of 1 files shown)


penelopeysm:
Hi @gragusa thanks for this PR! This is very helpful.

Could I ask you to add a test for this please? I think the most sensible place is in `test/display_tests.jl`. The method you modified targets `ChainDataFrame`, so I think just making one of those (e.g. via `mean(chain)`) and testing that `display(...)` doesn't error should be fine.

src/summarize.jl
```diff
@@ -25,15 +25,33 @@ end
...
-    formatter = PrettyTables.ft_printf("%.$(digits)f")
-
-    println(io, df.name)
-    # Support for PrettyTables 0.9 (`borderless`) and 0.10 (`tf_borderless`)
-    PrettyTables.pretty_table(
-        io, df.nt;
-        formatters = formatter,
-        tf = isdefined(PrettyTables, :borderless) ? PrettyTables.borderless : PrettyTables.tf_borderless,
-    )
+    if pkgversion(PrettyTables) < v"3.0"
```
penelopeysm, line 28:
```suggestion
    @static if pkgversion(PrettyTables) < v"3.0"
```

I'm not sure it really makes a difference, but it probably can't hurt.

devmotion: `pkgversion` is not supported on Julia < 1.9. IMO the easiest (for the compiler and for maintenance) would be to just require PrettyTables 3 and not introduce any code that is conditional on the loaded PrettyTables version.

gragusa: Requiring PrettyTable 3.0 would be great. This is just a transition. I am writing a package that uses both Prettytable 3.0 and MCMCChains, which is currently impossible. Switching to PrettyTable 3.0, on the other hand, would make people using PrettyTable 2.0 in their packages unhappy. So, this is a middle-ground solution. 

Could `pkgversion` be "backported"? The code is relatively easy and could be reimplemented for earlier versions of Julia.

penelopeysm: I don't think anything can be backported to < Julia 1.10 at this stage.

> Switching to PrettyTable 3.0, on the other hand, would make people using PrettyTable 2.0 in their packages unhappy.

I think it's OK, since there will still be an old version of MCMCChains that uses 2.0, and can be downloaded from the registry.

penelopeysm:
OK, somewhat ironically, PrettyTables@3 only works on Julia 1.10+.

src/summarize.jl
pkofod: This is causing issues for me right now unfortunately, because pkgversion doesn't work in our systemimage build   because pkgversion returns nothing.

penelopeysm: Okay, that's quite unfortunate. I guess we can change it to an isdefined check on something in PrettyTables v3.

devmotion: Ref https://github.com/JuliaLang/julia/issues/60085

devmotion: Ref https://github.com/JuliaLang/julia/issues/60085

penelopeysm: @pkofod I just triggered a release of 7.7.0, which should be out soon.

